PR #4, opened 2022-01-12
DT-3033 Promote to Prod

Mjwillis:


xoen (approved):

helm_deploy/values-prod.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+generic-service:
+  replicaCount: 4
+
+  ingress:
+    host: incentives-api.hmpps.service.justice.gov.uk
+    contextColour: green
+
+  env:
+    APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
```
petergphillips, line 12: shouldn't need this variable - it is the default in the `values.yaml` anyway

helm_deploy/values-prod.yaml
```diff
@@ -0,0 +1,17 @@
+---
+# Per environment values which override defaults in hmpps-incentives-api/values.yaml
+
+generic-service:
+  replicaCount: 4
+
+  ingress:
+    host: incentives-api.hmpps.service.justice.gov.uk
+    contextColour: green
```
petergphillips, line 9: would have been better putting `contextColour` in values.yaml and then you wouldn't need to set here

helm_deploy/values-prod.yaml
```diff
@@ -0,0 +1,17 @@
+---
+# Per environment values which override defaults in hmpps-incentives-api/values.yaml
+
+generic-service:
+  replicaCount: 4
```
petergphillips, line 5: same as the default value anyway so not necessary

Mjwillis: 👍

helm_deploy/values-prod.yaml
Mjwillis: 👍

PR #8, opened 2022-01-20
INC-97 First pass on Incentive Behaviour data

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/AuthTokenFilterFunction.kt
```diff
@@ -28,3 +30,12 @@ class AuthTokenResponseMono(
     next.exchange(mutatedRequest).subscribe(subscriber)
   }
 }
+
+fun <T> Flux<T>.withToken(authToken: String): Flux<T> =
+  this.contextWrite { transferAuthHeader(authToken, it) }
```
andymarke, line 35: Probably worth seeing if you can avoid having to make the additional `withToken` call.
I'm I sure I saw an example where they somehow add a auth token to the context in a filter somehow.

It is a common use  case so I bet it can be done.

andymarke (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/AuthTokenFilterFunction.kt
Mjwillis: indeed

PR #16, opened 2022-01-27
INC-144 Restore IEP level code to response

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/data/BehaviourSummary.kt
```diff
@@ -56,7 +77,7 @@ data class PrisonerIncentiveSummary(
   )
   val bookingId: Long,
   @Schema(description = "Internal reference for looking up the prisoners latest photo", example = "112121")
-  val imageId: Long,
+  val imageId: Long?,
```
xoen, line 80: Oh, does this mean the `imageId` is optional then?

xoen (approved):
Had a look and FWIW (given I don't know Kotlin) LGTM

But don't know if you're happy with my uneducated review or prefer someone else with Kotlin experience to have a look 🤷

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/data/BehaviourSummary.kt
Mjwillis: yes

PR #54, opened 2022-03-24
INC-297 Add audit and domain event

Mjwillis:


docker-compose-local.yaml
```diff
@@ -0,0 +1,39 @@
+version: "3"
```
neilmendum, line 1: Could name this `docker-compose-infra` because it runs both localstack and the postgres db

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelResourceTest.kt
```diff
@@ -3,11 +3,11 @@ package uk.gov.justice.digital.hmpps.incentivesapi.resource
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.IepReview
-import uk.gov.justice.digital.hmpps.incentivesapi.integration.IntegrationTestBase
+import uk.gov.justice.digital.hmpps.incentivesapi.integration.SqsIntegrationTestBase
 import java.time.LocalDate.now
 import java.time.format.DateTimeFormatter
 
-class IepLevelResourceTest : IntegrationTestBase() {
+class IepLevelResourceTest : SqsIntegrationTestBase() {
```
neilmendum, line 10: Could be worth verifying that `snsService` and `auditService` have been invoked as expected for `addIepReview` functions

Mjwillis: yes was going to do in next PR

docker-compose-local.yaml
Mjwillis: was following other conversions on the name

PR #58, opened 2022-04-04
INC-293: use same Prison API endpoint which DPS used to call

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-293

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@ services:
     depends_on:
       - hmpps-auth
     ports:
-      - "8081:8080"
+      - "8080:8080"
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:8080/health/ping"]
     environment:
@@ -20,7 +20,7 @@ services:
     image: quay.io/hmpps/hmpps-auth:latest
     networks:
       - hmpps
-    container_name: hmpps-auth
+    container_name: hmpps-auth-incentives-api
     ports:
       - "9090:8080"
     healthcheck:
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -27,10 +27,7 @@ class PrisonerIepLevelReviewService(
 ) {
   suspend fun getPrisonerIepLevelHistory(bookingId: Long, useNomisData: Boolean = true, withDetails: Boolean = true): IepSummary {
     return if (useNomisData) {
-      val result = prisonApiService.getIEPSummaryPerPrisoner(listOf(bookingId)).first()
```
neilmendum, line 30: When calling `POST:/bookings/iepSummary` it passes `withDetails=true` to [bookingService.getBookingIEPSummary](https://github.com/ministryofjustice/prison-api/blob/82a0d1a851a8221633babad51bfa21c24c8a4615/src/main/java/uk/gov/justice/hmpps/prison/api/resource/BookingResource.java#L512-L522). Prison API called `verifyBookingAccess` for the booking ID which resulted in a 404 because the DPS user did not have booking access. This caused Incentives API to return [500s in production](https://mojdt.slack.com/archives/CP8FR9XL5/p1649079873426879)

That DPS user has `ROLE_GLOBAL_SEARCH` and so `!isAllowedToViewAllPrisonerData` is false. Therefore if Incentives API uses the same Prison API endpoint which DPS used to uses and passes `withDetails` then the quick view screen will again work for all users

docker-compose.yml
```diff
@@ -50,11 +50,11 @@ services:
     image: quay.io/hmpps/prison-api:latest
     networks:
       - hmpps
-    container_name: prison-api
+    container_name: prison-api-incentives-api
     depends_on:
       - hmpps-auth
     ports:
-      - "8080:8080"
+      - "8082:8080"
```
neilmendum, line 57: The port configured in `application-dev.yml`

Mjwillis (approved):
LGTM

PR #59, opened 2022-04-06
INC-471 Prisons with Entry IEP are causing an error when loading IEP Levels

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
```diff
@@ -19,13 +20,19 @@ class IepLevelService(
   @Transactional(readOnly = true)
   suspend fun getIepLevelsForPrison(prisonId: String): List<IepLevel> {
     return coroutineScope {
-      val iepLevelMap = iepLevelRepository.findAllByActiveIsTrueOrderBySequence().toList().associateBy { it.iepCode }
+      val iepLevelMap = iepLevelRepository.findAll().toList().associateBy { it.iepCode }
```
xoen, line 23: Does this change the order of the IEP levels?

xoen (approved):

PR #131, opened 2022-07-28
INC-566: return HTTP status 400 if it receives undefined parameter

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-566

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelResourceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelResourceTest.kt
@@ -98,6 +98,14 @@ class IepLevelResourceTest : SqsIntegrationTestBase() {
   }
 
   @Test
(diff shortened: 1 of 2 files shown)


andrewrlee (approved):
lgtm

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelResourceTest.kt
```diff
@@ -97,6 +97,14 @@ class IepLevelResourceTest : SqsIntegrationTestBase() {
       )
   }
 
+  @Test
+  fun `handle undefined path variable`() {
+    webTestClient.get().uri("/iep/reviews/booking/undefined")
```
andrewrlee, line 102: presumably as node passes `undefined` as a string due to dodgy concat, and this requires a long?

neilmendum: Bingo!

PR #138, opened 2022-08-15
INC-697: pass includeInactive=true to prison-api getLocation

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-697


xoen (approved):
FWIW LGTM

PR #143, opened 2022-08-17
INC-672: Implement the transfer policy logic

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-672

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
@@ -2,7 +2,6 @@ package uk.gov.justice.digital.hmpps.incentivesapi.service
 
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.first
-import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.toList
 import org.slf4j.Logger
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -122,6 +122,28 @@ class PrisonerIepLevelReviewService(
...
 
+  private suspend fun getIepLevelForReviewType(prisonerInfo: PrisonerAtLocation, reviewType: ReviewType): String {
+    val iepLevelsForPrison = iepLevelService.getIepLevelsForPrison(prisonerInfo.agencyId)
+    when (reviewType) {
+      ReviewType.INITIAL -> {
+        return iepLevelsForPrison.first(IepLevel::default).iepLevel
+      }
+
+      ReviewType.TRANSFER -> {
+        val currentIepLevelForPrisoner = getCurrentIEPLevelForPrisoners(bookingIds = listOf(prisonerInfo.bookingId), useNomisData = true, useClientCredentials = true)
```
xoen, line 133: Do we know if at the point of the this `TRANSFER` message being processed the current prisoner IEP level is the one at the old prison (before transfer) or if it's already set to the new prison default somehow?

Don't know if my question makes sense.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -122,6 +122,28 @@ class PrisonerIepLevelReviewService(
...
+          .firstOrNull() ?: throw NoDataFoundException(prisonerInfo.bookingId)
+        return iepLevelsForPrison.find { it.iepDescription == currentIepLevelForPrisoner.iepLevel }
+          ?.iepLevel
+          ?: run {
```
xoen, line 137: is this `?: run` some kind of ternary operator (if/else)?

I'm guessing this means:

> if we can find the current prisoner level in the prison levels
>    => return this level
> else
>    => return the highest level for the prison

Is that correct? Sorry this is a bit terse, I should have a look at the Kotlin syntax.

neilmendum: Exactly. It's the [Elvis operator](https://kotlinlang.org/docs/null-safety.html#elvis-operator). `find` could return null and so we want to handle that scenario with the logic you described in the quote block

I did think about extracting `iepLevelsForPrison.find { it.iepDescription == currentIepLevelForPrisoner.iepLevel }` into a variable - something like `prisonSupportsPrisonerIepLevel` - but wasn't sure if it would help with readability

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
neilmendum: That is a really good point. I think we need to do some digging into Prison API for a transferred prisoner to see how the incentive level history looks

neilmendum: [Writeup of investigation](https://dsdmoj.atlassian.net/browse/INC-672?focusedCommentId=235075)

xoen (dismissed):

xoen (approved):

PR #146, opened 2022-08-23
Improved SNS instructions in README

xoen:
It is possible that publishing an error to SNS locally fails if the
default AWS profile points to a region different than `eu-west-2`


neilmendum (approved):
LGTM

PR #149, opened 2022-08-26
INC-733: Set IEP level review time to when the domain event occurred

xoen:
The domain event contains the date/time of when the even occurred so
we can use this time in the created `PrisonerIepLevel` record instead
of "now" (when the domain event is processed).

Also:
- updated SNS message payload so that datetimes have same format as per production messages (with local time offset)
- simplified `filter {}` expression to get list of active prison levels

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
@@ -24,7 +24,7 @@ class IepLevelService(
 
       withContext(Dispatchers.Default) {
         iepPrisonRepository.findAllByPrisonIdAndActiveIsTrue(prisonId)
-          .filter { iepLevelMap[it.iepCode] != null && iepLevelMap[it.iepCode]!!.active }
+          .filter { iepLevelMap[it.iepCode]?.active == true }
           .map {
             val iepLevel = iepLevelMap[it.iepCode]!!
             IepLevel(
--- a/README.md
+++ b/README.md
@@ -49,8 +49,8 @@ aws --endpoint-url=http://localhost:4566 sns publish \
     }' \
     --message '{
       "version":"1.0",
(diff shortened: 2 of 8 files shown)


neilmendum (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
```diff
@@ -24,7 +24,7 @@ class IepLevelService(
 
       withContext(Dispatchers.Default) {
         iepPrisonRepository.findAllByPrisonIdAndActiveIsTrue(prisonId)
-          .filter { iepLevelMap[it.iepCode] != null && iepLevelMap[it.iepCode]!!.active }
```
xoen, line 27: @neilmendum it's interesting despite the `iepLevelMap[it.iepCode] != null &&` the Kotlin compiler still doesn't seem to infer that `iepLevelMap[it.iepCode]` is now a `Boolean` (not a `Boolean?`).

We should definitely [try and avoid using `!!`](https://kotlinlang.org/docs/null-safety.html#the-operator) when an alternative is available as we can avoid these `NullPointerException` this way (I guess I'm not one of those "NPE lovers" :P)

xoen: PS: I'll update the test to add a case for when a prison have some "weird" level which is not present in the global `IepLevel` table to cover the `iepLevelMap[it.iepCode] == null` case.

xoen: New test: https://github.com/ministryofjustice/hmpps-incentives-api/pull/149/files#diff-fc67634b9869e5d81f9583cbfd10674aaa100a11d0e83b97fcae6dc57e844dc1

neilmendum (approved):
LGTM

PR #154, opened 2022-09-06
INC-735: Avoid 500s when Prison API responds 404

xoen:
### Background
We saw some odd `NoSuchElementException` exceptions when Prison API responded with a `404 Not Found`, this meant these requests to the Incentive API responded with a `500 Internal Server Error`.

### Reason
A `NoSuchElementException` was caused by [calling `first()` on an empty flow](https://github.com/ministryofjustice/hmpps-incentives-api/blob/88481925a2360cfc8bbdc67bbb423e3338d9364e/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt#L56).

The `PrisonApiService` methods were swallowing the `NotFound` exceptions and returning an empty flow in those cases causing the `NoSuchElementException` above when calling `.first()` on it.

### Solution
`PrisonApiService` doesn't swallow exceptions anymore and `NotFound` exceptions are now handled and clients gets a `404 Not Found` instead of a `500` error.

### Ticket
https://dsdmoj.atlassian.net/browse/INC-735

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelResourceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelResourceTest.kt
@@ -43,6 +43,18 @@ class IepLevelResourceTest : SqsIntegrationTestBase() {
   }
 
   @Test
+  fun `Prison API '404 Not Found' responses are handled instead of responding 500 Internal Server Error`() {
+    val bookingId: Long = 1234134
+
+    prisonApiMockServer.stubApi404for("/api/bookings/$bookingId/iepSummary?withDetails=true")
+
+    webTestClient.get().uri("/iep/reviews/booking/$bookingId")
+      .headers(setAuthorisation())
+      .exchange()
+      .expectStatus().isNotFound
+  }
+
+  @Test
   fun `get IEP Levels for a prison`() {
 
     webTestClient.get().uri("/iep/levels/MDI")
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelResourceTest.kt
```diff
@@ -42,6 +42,18 @@ class IepLevelResourceTest : SqsIntegrationTestBase() {
       .isUnauthorized
   }
 
+  @Test
+  fun `Prison API '404 Not Found' responses are handled instead of responding 500 Internal Server Error`() {
+    val bookingId: Long = 1234134
+
+    prisonApiMockServer.stubApi404for("/api/bookings/$bookingId/iepSummary?withDetails=true")
+
+    webTestClient.get().uri("/iep/reviews/booking/$bookingId")
```
neilmendum, line 51: You could pass `use-nomis-data=true` so it's a bit clearer we're making a call to PrisonAPI?

xoen: True, how wonder why the test works tho 🤔 - I'll change and also have a play

neilmendum: [By default](https://github.com/ministryofjustice/hmpps-incentives-api/blob/aa445fea20258a2b7c8776e927e362a0b3749f5a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelResource.kt#L107) `use-nomis-data` is set to true - I just thought we could include it in the test for clarity but not a big issue 😄

xoen: yeah, just realised it was the default (as it makes sense) but it's better to be explicit 👍

xoen: https://github.com/ministryofjustice/hmpps-incentives-api/pull/154/commits/f0431201be252ad871fc22d010d47fb036d33160

neilmendum (approved):
LGTM

PR #155, opened 2022-09-09
Simplify suspend functions/codebase

xoen:
- removed unnecessary `coroutineScope`
- simplified `getIncentivesSummaryByLocation()` method
- use `fold()` when counting case notes
- use property references when it makes code a bit more concise
- moved BehaviourSummary classes in `dto` packages for consistency
- removed unnecessary `let {}`/`with()`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
@@ -7,9 +7,9 @@ import kotlinx.coroutines.flow.toList
 import org.apache.commons.text.WordUtils
 import org.springframework.data.domain.Sort
 import org.springframework.stereotype.Service
-import uk.gov.justice.digital.hmpps.incentivesapi.data.BehaviourSummary
-import uk.gov.justice.digital.hmpps.incentivesapi.data.IncentiveLevelSummary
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
```diff
@@ -146,11 +146,7 @@ class IncentiveSummaryService(
       }
 
   private fun calcTypeCount(caseNoteUsage: List<CaseNoteUsage>): Int =
-    if (caseNoteUsage.isNotEmpty()) {
-      caseNoteUsage.map { it.numCaseNotes }.reduce { acc, next -> acc + next }
-    } else {
-      0
-    }
+    caseNoteUsage.map { it.numCaseNotes }.fold(0) { acc, next -> acc + next }
```
neilmendum, line 143: Neat!

neilmendum (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
xoen: yeah, I prefer this because you don't have to deal with the empty list case as an exceptional one, works nice as `0` is the identity of the sum ☺️

PR #158, opened 2022-09-14
INC-736: raise IEP domain event for transfer and first arrival

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-736

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
@@ -30,6 +30,8 @@ class PrisonerIepLevelReviewService(
   private val prisonerIepLevelRepository: PrisonerIepLevelRepository,
   private val iepLevelRepository: IepLevelRepository,
   private val iepLevelService: IepLevelService,
+  private val snsService: SnsService,
+  private val auditService: AuditService,
   private val authenticationFacade: AuthenticationFacade,
   private val clock: Clock,
 ) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -240,6 +244,16 @@ class PrisonerIepLevelReviewService(
     )
   }
 
+  private suspend fun sendEventAndAudit(iepDetail: IepDetail) {
+    snsService.sendIepReviewEvent(iepDetail.id!!, iepDetail.iepTime)
```
xoen, line 248: I do wonder how a record could **_not_** have an ID just after the repository has saved it?

I'm wary of these `!!` because they could raise a `NullPointerException`, maybe the compiler was complaining about `id` being a `String?` instead of a `String`?

I suppose this is more of a very very exceptional case (e.g. shouldn't really happen) - however if it is kinda possible for the ID to be `null` after the repository `save()` call and assuming an exception wasn't raised by `save()` itself, it may be worth checking the `null` case explicitly and log a nice message and/or raise a custom exception...

...but I think in this case the result of `save()` should "always" have an `id` right?

xoen (dismissed):

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewServiceTest.kt
```diff
@@ -112,6 +118,14 @@ class PrisonerIepLevelReviewServiceTest {
 
   @Nested
   inner class processReceivedPrisoner {
+    @BeforeEach
+    fun setUp(): Unit = runBlocking {
+      // This ensures save works and an id is set on the PrisonerIepLevel
+      whenever(prisonerIepLevelRepository.save(any())).thenAnswer { i -> i.arguments[0] }
+      whenever(iepLevelRepository.findById("STD")).thenReturn(uk.gov.justice.digital.hmpps.incentivesapi.jpa.IepLevel(iepCode = "STD", iepDescription = "Standard"))
```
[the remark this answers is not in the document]
neilmendum: Good spot 👍 I must have pressed the wrong button in IntelliJ when importing it. Have imported it correctly now and dropped the need to fully qualify it

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
neilmendum: Good point! Have wrapped that in `let`

xoen (approved):

PR #161, opened 2022-09-15
INC-737: Remove `sequence` from model (BREAKING change)

xoen:
[Removed `sequence` from `prisoner_iep_level` DB table](https://github.com/ministryofjustice/hmpps-incentives-api/commit/9179145491fe7f7c3216cc19d8fda7c262dfcfb7) and corresponding API responses.

**NOTE**: [`IepLevel` has a `sequence` column too](https://github.com/ministryofjustice/hmpps-incentives-api/blob/adf4f6d000ec0781685370f2df66172799b2c0e9/src/main/resources/db/migration/V1_3__iep_level_data.sql) but that is used to sort IEP levels (e.g. `Basic` < `Standard`). This is **not** removed.
 

⚠️ BREAKING change

PR #166, opened 2022-09-20
INC-769: handle null userId when migrating record from NOMIS

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-769

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
@@ -86,7 +86,7 @@ class PrisonerIepLevelReviewService(
         locationId = iepMigration.locationId,
         sequence = 0,
         current = iepMigration.current,
-        reviewedBy = iepMigration.userId ?: "NOMIS",
+        reviewedBy = iepMigration.userId,
         reviewTime = iepMigration.iepTime,
         reviewType = iepMigration.reviewType,
         prisonerNumber = prisonerInfo.offenderNo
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -86,7 +86,7 @@ class PrisonerIepLevelReviewService(
         locationId = iepMigration.locationId,
         sequence = 0,
         current = iepMigration.current,
-        reviewedBy = iepMigration.userId,
+        reviewedBy = iepMigration.userId ?: "NOMIS",
```
xoen, line 89: I don't know how I feel about this 🤔 

I see the value of avoiding a `null` value, but the issue with this is that field values are for user IDs, so the client may make a request to get user information (e.g. to show the user full name or something) - and in this case it would not work.

(Unless there is a special user with ID `"NOMIS"` used across DPS for these kind of situations I'm not aware of)

The clients may have special logic to handle this special user ID but it somehow doesn't feel right 🤔

ushkarev: i would have thought we'd keep it `null` (which is true to NOMIS) and simply choose to present that in some way in DPS. "NOMIS" would not necessarily be meaningful to users of DPS and i'm sure the team would want involvement in choosing the right word

neilmendum: Very good points - have changed `prisoner_iep_level.reviewedBy` to be nullable and removed setting it to `NOMIS` during the migration flow

src/main/resources/db/migration/V1_12__prisoner_iep_nullable_reviewed_by.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE prisoner_iep_level ALTER COLUMN reviewed_by DROP NOT NULL;
```
neilmendum, line 1: Named this `V1_12` because in https://github.com/ministryofjustice/hmpps-incentives-api/pull/161 we have `V1_11` and wanted to avoid naming conflicts as well as Flyway confusion

xoen: Thanks :) (I'll probably merge that PR later this morning anyway 👍 )

xoen (approved):

PR #170, opened 2022-09-26
INC-738: POST /iep/sync/booking/{bookingId} endpoint

xoen:
Endpoint invoked to synchronise IEP reviews from NOMIS.

As per ticket this endpoint has same request format as per migration endpoint (but it publishes the domain events).
As part of this I've renamed the `IepMigration` data class into `SyncPostRequest`.


neilmendum (approved):
LGTM

PR #172, opened 2022-09-28
INC-765: add iepCode to IepDetail

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-765


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
```diff
@@ -65,6 +65,8 @@ data class IepDetail(
   val id: Long? = null,
   @Schema(description = "IEP Level", example = "Standard", required = true)
   val iepLevel: String,
+  @Schema(description = "IEP Code", example = "STD", required = false)
```
neilmendum, line 68: Made this optional because we also use `IepSummary` as the [return type from PrisonAPI](https://github.com/ministryofjustice/hmpps-incentives-api/blob/34319de9edf22b0a806125d739852df15edf40a8/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonApiService.kt#L25-L39) and PrisonAPI does not provide it

I did think about making it required, but this would involve

1. Introducing a new type for PrisonAPI responses
2. Mapping the `iepLevel` returned to `iepCode`

Happy to discuss if folks think the alternative is better!

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -161,7 +161,7 @@ class PrisonerIepLevelReviewService(
         val iepHistory = getPrisonerIepLevelHistory(prisonerInfo.bookingId, useNomisData = true, withDetails = true, useClientCredentials = true).iepDetails
         val iepLevelBeforeTransfer = iepHistory.sortedBy(IepDetail::iepTime).lastOrNull { it.agencyId != prisonerInfo.agencyId }
           ?: throw NoDataFoundException(prisonerInfo.bookingId)
-        iepLevelsForPrison.find { it.iepLevel == iepLevelBeforeTransfer.iepLevel }
+        iepLevelsForPrison.find { it.iepDescription == iepLevelBeforeTransfer.iepLevel }
```
neilmendum, line 164: When updating the test `PrisonerIepLevelReviewServiceTest.process TRANSFERRED when offender stays on same incentive level as previous prison` I realised we were setting `IepDetail.iepLevel` to the abbreviated format of an IEP level. For example `STD` rather than `Standard`

We should be comparing either
`IepLevel.iepDescription` with `IepDetail.iepLevel`
or
`IepLevel.iepLevel` with `IepDetail.iepCode`

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -269,12 +269,13 @@ class PrisonerIepLevelReviewService(
...
       agencyId = prisonId,
       iepLevel = iepLevelRepository.findById(iepCode)?.iepDescription ?: "Unmapped",
+      iepCode = iepCode,
       reviewType = reviewType,
       comments = commentText,
       userId = reviewedBy,
       locationId = locationId,
       prisonerNumber = prisonerNumber,
-      auditModuleName = "Incentives-API"
+      auditModuleName = "Incentives-API",
```
xoen, line 278: trailing commas 😍

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
xoen: This makes sense.

However [the `IepLevel` class](https://github.com/ministryofjustice/hmpps-incentives-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonApiDto.kt#L42-L51) is a bit evil here as its fields are named differently and caused this understandable confusion:

```kotlin
data class IepLevel(
  @Schema(description = "IEP Code for an IEP level", example = "STD")
  val iepLevel: String,
  @Schema(description = "Description of the IEP Level", example = "Standard")
  val iepDescription: String,
  // [...]
```

In other places `iepLevel` is the description (e.g. `"Standard"`) and `iepCode` is the code (e.g. `"STD"`)...but not in that class :(

I don't think that class is exposed in any of our API responses. I can't quite tell if it's used to deserialise a response from Prison API.

I think it would be ideal to update this class to make it consistent to the others and make the `iepLevel` the description and `iepCode` the code (if it doesn't break compatibility of our API/Prison API responses)

xoen: OK, I was playing with this locally and it seems [there is at least one place in the API we use this `IepLevel` class to construct the response (`GET /iep/levels/{prisonId}`)](https://github.com/ministryofjustice/hmpps-incentives-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelResource.kt#L39-L76), so no, we cannot rename these fields without potentially breaking compatibility with some of the consumers 🤷

xoen (approved):
Looks great 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
ushkarev: for what it's worth, DPS uses `iepLevel` fields (with all their different meanings) from several endpoints so we're kinda stuck with this mixup now

PR #184, opened 2022-10-04
INC-822: use client credentials when making calls to PrisonAPI during sync and migration flows

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-822

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonApiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonApiService.kt
@@ -74,11 +74,13 @@ class PrisonApiService(
       .awaitBody()
   }
 
-  suspend fun getPrisonerInfo(bookingId: Long): PrisonerAtLocation =
-    prisonWebClient.get()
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonApiService.kt
```diff
@@ -74,11 +74,13 @@ class PrisonApiService(
       .awaitBody()
   }
 
-  suspend fun getPrisonerInfo(bookingId: Long): PrisonerAtLocation =
-    prisonWebClient.get()
+  suspend fun getPrisonerInfo(bookingId: Long, useClientCredentials: Boolean = false): PrisonerAtLocation {
```
neilmendum, line 77: This is called by other flows, for example from DPS when it POSTs to `/iep/reviews/prisoner/{prisonerNumber}` and that call uses `prisonWebClient` which propagates the `hmpps-incentives-ui` client credentials. For migration and sync we're changing it to use the service's client credentials [provided in the namespace](https://github.com/ministryofjustice/hmpps-incentives-api/blob/f8e4563969640d393ccc5e81e0dce3715c9afe60/helm_deploy/hmpps-incentives-api/values.yaml#L45-L46)

PR #187, opened 2022-10-05
INC-815: Basic next review date calculation

ushkarev:
• Adds new derived field to `IepSummary` with a very simple calculation for next review date. `lastReviewDate + 1 year` applies to people not on basic, not newly in prison, not recalled and not transferred since last review.
• Fixes some minor test issues
• Changes how `daysSinceReview` is calculated to prevent errors and confusion, see explanation in 273f44ae3c934a62e8a5b69c5758bd75cf608c11

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
@@ -39,6 +39,10 @@ data class IepSummary(
       return Duration.between(iepDate.atStartOfDay(), currentIepDate).toDays().toInt()
     }
 
+  @get:Schema(description = "Date of next review", example = "2022-12-31", required = true)
+  @get:JsonProperty
+  val nextReviewDate: LocalDate = iepDate.plusYears(1)
+
   fun daysOnLevel(): Int {
     val currentIepDate = LocalDate.now().atStartOfDay()
     var daysOnLevel = 0
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.incentivesapi.dto
 
 import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonProperty
 import io.swagger.v3.oas.annotations.media.Schema
 import uk.gov.justice.digital.hmpps.incentivesapi.jpa.ReviewType
 import java.time.Duration
@@ -24,17 +25,19 @@ data class IepSummary(
   val iepDate: LocalDate,
   @Schema(description = "Date and time when last review took place", required = true, example = "2021-12-31T12:34:56.789012")
   val iepTime: LocalDateTime,
-  @Schema(description = "Days Since last Review", example = "23", required = true)
-  val daysSinceReview: Int,
   @Schema(description = "Location  of prisoner when review took place within prison (i.e. their cell)", example = "1-2-003", required = false)
   val locationId: String? = null,
   @Schema(description = "IEP Review History (descending in time)", required = true)
   val iepDetails: List<IepDetail>,
 ) {
 
-  fun daysSinceReview(): Int {
-    return daysSinceReview(iepDetails)
-  }
+  @get:Schema(description = "Days since last review", example = "23", required = true)
+  @get:JsonProperty
+  val daysSinceReview: Int
+    get() {
+      val currentIepDate = LocalDate.now().atStartOfDay()
+      return Duration.between(iepDate.atStartOfDay(), currentIepDate).toDays().toInt()
+    }
 
   fun daysOnLevel(): Int {
     val currentIepDate = LocalDate.now().atStartOfDay()
@@ -51,11 +54,6 @@ data class IepSummary(
 
     return daysOnLevel
   }
-}
-
-fun daysSinceReview(iepHistory: List<IepDetail>): Int {
-  val currentIepDate = LocalDate.now().atStartOfDay()
-  return Duration.between(iepHistory.first().iepDate.atStartOfDay(), currentIepDate).toDays().toInt()
 }
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
@@ -35,8 +35,8 @@ data class IepSummary(
   @get:JsonProperty
   val daysSinceReview: Int
     get() {
-      val currentIepDate = LocalDate.now().atStartOfDay()
-      return Duration.between(iepDate.atStartOfDay(), currentIepDate).toDays().toInt()
+      val today = LocalDate.now().atStartOfDay()
+      return Duration.between(iepDate.atStartOfDay(), today).toDays().toInt()
     }
 
   @get:Schema(description = "Date of next review", example = "2022-12-31", required = true)
@@ -44,7 +44,7 @@ data class IepSummary(
   val nextReviewDate: LocalDate = iepDate.plusYears(1)
 
   fun daysOnLevel(): Int {
-    val currentIepDate = LocalDate.now().atStartOfDay()
+    val today = LocalDate.now().atStartOfDay()
     var daysOnLevel = 0
 
     run iepCheck@{
@@ -52,7 +52,7 @@ data class IepSummary(
         if (it.iepLevel != iepLevel) {
           return@iepCheck
         }
-        daysOnLevel = Duration.between(it.iepDate.atStartOfDay(), currentIepDate).toDays().toInt()
+        daysOnLevel = Duration.between(it.iepDate.atStartOfDay(), today).toDays().toInt()
       }
     }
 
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
```diff
@@ -53,11 +56,6 @@ data class IepSummary(
   }
 }
 
-fun daysSinceReview(iepHistory: List<IepDetail>): Int {
-  val currentIepDate = LocalDate.now().atStartOfDay()
-  return Duration.between(iepHistory.first().iepDate.atStartOfDay(), currentIepDate).toDays().toInt()
```
ushkarev, line 58: `iepHistory` is an empty list when `IepSummary` is loaded without details: this would have thrown an exception. as would the method, but not the field!

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewServiceTest.kt
```diff
@@ -549,7 +549,6 @@ class PrisonerIepLevelReviewServiceTest {
   private val iepTime: LocalDateTime = LocalDateTime.now().minusDays(10)
   private fun iepSummary(iepLevel: String = "Enhanced", iepDetails: List<IepDetail> = emptyList()) = IepSummary(
     bookingId = 1L,
-    daysSinceReview = 60,
```
ushkarev, line 552: `daysSinceReview` **is** in fact provided by prison api. it can't be here as is because this mock already presumes that prison api's JSON has been deserialised.

question: does the deserialisation system ignore superfluous/unexpected fields?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
```diff
@@ -24,17 +25,19 @@ data class IepSummary(
...
   @Schema(description = "IEP Review History (descending in time)", required = true)
   val iepDetails: List<IepDetail>,
 ) {
 
-  fun daysSinceReview(): Int {
-    return daysSinceReview(iepDetails)
-  }
+  @get:Schema(description = "Days since last review", example = "23", required = true)
+  @get:JsonProperty
+  val daysSinceReview: Int
```
ushkarev, line 36: `daysSinceReview` is a derived property of `iepDate` so should always be calculated from that. we can trust both what prison api returned and what we calculate from the history to make the _field_ correct, however the _method_ would have thrown an exception if the instance was built without details.

also, forcing this to be a derived property means that it's impossible to pass inconsistent data into the constructor.

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/integration/wiremock/PrisonApiMockServer.kt
```diff
@@ -281,15 +281,15 @@ class PrisonApiMockServer : WireMockServer(WIREMOCK_PORT) {
           .withBody(
             """
               {
-                "bookingId": 1234134,
+                "bookingId": $bookingId,
```
ushkarev, line 284: since the requested `bookingId` can vary, it makes sense to respond with a matching one!
NB: no tests relied on past (arguably incorrect) functionality

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewServiceTest.kt
ushkarev: seems the answer is that the superfluous/unexpected field _is_ ignored as proved by tests using this stub still passing:
https://github.com/ministryofjustice/hmpps-incentives-api/blob/f8e4563969640d393ccc5e81e0dce3715c9afe60/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/integration/wiremock/PrisonApiMockServer.kt#L141

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
```diff
@@ -24,17 +25,19 @@ data class IepSummary(
...
+    get() {
+      val currentIepDate = LocalDate.now().atStartOfDay()
```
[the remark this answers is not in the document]
xoen: `iepDate` field is really the lastest/current IEP date AFAICT right?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
xoen: I didn't check, but when getting this without details, is the `iepDate` field just set to the latest IEP date somehow?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
```diff
@@ -24,17 +25,19 @@ data class IepSummary(
...
   val iepDetails: List<IepDetail>,
 ) {
 
-  fun daysSinceReview(): Int {
-    return daysSinceReview(iepDetails)
-  }
+  @get:Schema(description = "Days since last review", example = "23", required = true)
+  @get:JsonProperty
+  val daysSinceReview: Int
+    get() {
```
xoen, line 38: How is this `get() { }` works? Is it a getter for the `val` immediately above?

And presumably this is why you can now replace `.daysSinceReview()` calls with just `.daysSinceReview`?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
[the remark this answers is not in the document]
ushkarev: good idea, will rename… i just copied what was there
https://github.com/ministryofjustice/hmpps-incentives-api/blob/423f0afedf854b62fda64c2a8cf23292f19e3b1f/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt#L57

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
```diff
@@ -39,21 +40,24 @@ data class IepSummary(
...
-    var daysOnLevel = 0
+  val daysOnLevel: Int
+    get() {
+      if (iepDetails.isEmpty()) throw ValidationException("iepDetails not available")
 
-    run iepCheck@{
-      iepDetails.forEach {
-        if (it.iepLevel != iepLevel) {
-          return@iepCheck
+      val currentIepDate = LocalDate.now().atStartOfDay()
```
xoen, line 51: I know this is the existing code wrapped in a getter...but similar to other comment, I think this "variable" should be called `today` to avoid any confusion ☺️

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
```diff
@@ -46,7 +46,7 @@ data class IepSummary(
 
   val daysOnLevel: Int
     get() {
-      if (iepDetails.isEmpty()) throw ValidationException("iepDetails not available")
+      if (iepDetails.isEmpty()) throw IncorrectUsage("iepDetails for booking $bookingId not loaded")
```
xoen, line 49: You seen, this is the problem with raising this exception now.

Exception suggests "Incorrect usage" but is it? The consumer of the API didn't do anything wrong, it just happened to make a request for which one some of the fields couldn't be calculated and it's getting a `400 Bad Request` (I haven't checked this `IncorrectUsage` exception is handled to response `400 Bad Request` but I'm assuming that would be the case).

xoen: Actually AFAICT [I don't see this exception being handled by Spring Boot](https://github.com/ministryofjustice/hmpps-incentives-api/blob/INC-815-basic-next-review-date/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/HmppsIncentivesApiExceptionHandler.kt), which should result in a `500 Internal Server Error` for the consumer.

I am still not sure we should raise an exception when we cannot calculate one of the fields ([see previous comment](https://github.com/ministryofjustice/hmpps-incentives-api/pull/187#discussion_r988921210))

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
```diff
@@ -24,38 +26,42 @@ data class IepSummary(
...
+  @get:Schema(description = "Date of next review", example = "2022-12-31", required = true)
+  @get:JsonProperty
+  val nextReviewDate: LocalDate = iepDate.plusYears(1)
 
-  fun daysOnLevel(): Int {
-    val currentIepDate = LocalDate.now().atStartOfDay()
-    var daysOnLevel = 0
+  val daysOnLevel: Int
+    get() {
+      if (iepDetails.isEmpty()) throw ValidationException("iepDetails not available")
```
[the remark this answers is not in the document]
ushkarev: durrrrrrr i was stupid. will _instead_ add a test to cover checking a response without details (which would otherwise have caught this stupidity)

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/DaysOnLevelTest.kt
```diff
@@ -117,9 +117,9 @@ class DaysOnLevelTest {
       val iepTime = LocalDateTime.now().minusDays(10)
       val iepSummary = IepSummary(
         bookingId = 1L,
-        iepDate = iepTime.toLocalDate(),
+        iepDate = iepTime.toLocalDate().plusDays(9),
```
xoen, line 120: I'm almost tempted to suggest building the list of `IepDetail` before the `IepSummary` and use `iepDetails[0].iepDate` to make this relationship even more explicit...

but I'm also happy with this :)

PR #188, opened 2022-10-06
INC-824: Fix handling of 'TRANSFERRED' domain event when 2+ current=true

xoen:
The `prisoner_iep_level` DB table has `current` boolean column (`current` is something of a legacy concept coming from NOMIS. In theory the record with latest `review_time` should be the "current")

More than one record can currently have that set to `true` which can cause a problem for the following method/generated query that expect only one result:

```
findOneByBookingIdAndCurrentIsTrue()
```

I've changed this method to `findFirstByBookingIdAndCurrentIsTrueOrderByReviewTimeDesc()`:
- so it sorts by `review_time` DESC (as that's the natural order)
- it returns the first record with `current=true`

This should avoid that exception to be raised on transfer.

I've also added ordering to the other repository method:
- renamed `findAllByBookingIdInAndCurrentIsTrue()`
- to `findAllByBookingIdInAndCurrentIsTrueOrderByReviewTimeDesc()`


Mjwillis (approved):
LGTM

PR #190, opened 2022-10-06
INC-827: Improve `daysOnLevel` calculation

ushkarev:
Previously, it silently returned 0 when history was unavailable (i.e. `IepSummary` was loaded "without details") instead of indicating that there was a problem. It was only used in one place (on an intermediate object) and was not serialised as part of the `IepSummary` schema.

Throwing an exception seems prudent to indicate that there is an _implementation_ problem (wrong details passed in). The calculation has also been moved closer to where it's used. It's less error prone than being a method that looked reusable.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
@@ -89,7 +89,7 @@ class IncentiveSummaryService(
       levels[prisonerMap.key] ?: invalidLevel()
     }
 
-  fun getPrisonersByLevel(
+  private fun getPrisonersByLevel(
     prisoners: List<PrisonerAtLocation>,
     prisonerLevels: Map<Long, IepResult>
   ): Map<String, List<PrisonerAtLocation>> =
@@ -97,7 +97,7 @@ class IncentiveSummaryService(
       prisonerLevels[it.bookingId]?.iepLevel ?: missingLevel().iepLevel
     }
 
-  fun addMissingLevels(
+  private fun addMissingLevels(
     data: List<IncentiveLevelSummary>,
     levelMap: Map<String, IepLevel>
   ): List<IncentiveLevelSummary> {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
```diff
@@ -159,6 +162,20 @@ data class IepResult(
   val daysOnLevel: Int
 )
 
+fun daysOnLevel(iepDetails: List<IepDetail>): Int {
+  val today = LocalDate.now().atStartOfDay()
+
+  val earliestMatchingIepDetail = iepDetails.reduce { earliestMatchingIepDetail, iepDetail ->
```
ushkarev, line 168: `reduce` throws an `UnsupportedOperationException` if the iterable is empty. it should be clear enough without needing a bespoke exception class.

a bespoke class would require this block to be nested in a try-catch that re-throws a different exception… kinda unnecessary overkill?

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepSummaryDateTest.kt
```diff
@@ -244,7 +245,24 @@ class IepSummaryDateTest {
...
+      val iepSummary = IepSummary(
+        bookingId = 1L,
+        iepDate = iepTime.toLocalDate(),
+        iepLevel = "Basic",
+        iepTime = iepTime,
+        iepDetails = emptyList(),
+      )
+
+      assertThat(iepSummary.daysSinceReview).isEqualTo(3)
+      assertThatThrownBy { daysOnLevel(iepSummary.iepDetails) }
```
ushkarev, line 263: the `iepSummary` object is not needed but it demonstrates how `daysOnLevel` is actually used in main code

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
xoen: yeah, if `reduce {}` already throw an exception when list is empty I think that's good enough, especially for a programming error that should never happen.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
```diff
@@ -159,6 +162,20 @@ data class IepResult(
   val daysOnLevel: Int
 )
 
+fun daysOnLevel(iepDetails: List<IepDetail>): Int {
```
xoen, line 165: can this `daysOnLevel` utility function be private? Or do the tests need this to be public to use it?

xoen (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
ushkarev: i tried that but the tests couldn't see it even though they're in the same package. not really sure what private even means when it's top-level in a package.

PR #194, opened 2022-10-07
INC-831: drop SyncPostRequest.locationId and make it nullable

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-826

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
@@ -76,15 +76,20 @@ class PrisonerIepLevelReviewService(
   }
 
   @Transactional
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -76,15 +76,20 @@ class PrisonerIepLevelReviewService(
   }
 
   @Transactional
-  suspend fun persistPrisonerIepLevel(bookingId: Long, syncPostRequest: SyncPostRequest): IepDetail {
+  suspend fun persistSyncPostRequest(bookingId: Long, syncPostRequest: SyncPostRequest, includeLocation: Boolean): IepDetail {
```
neilmendum, line 79: Renamed this because we also have `persistIepLevel` in the class and I was getting confused, happy to discuss

neilmendum:
Pushed against wrong branch name - d'oh. Raised https://github.com/ministryofjustice/hmpps-incentives-api/pull/195

PR #197, opened 2022-10-12
INC-760: Feature flag to either publish domain event or call Prison API on new IEP review

xoen:
### Current state
When a new IEP review is created two things happen to update NOMIS/notify other services that may be interested in this:
1. a domain event is published
2. an HTTP request to Prison API's `POST /api/bookings/{bookingId}/iepLevels` is made

A sync service is being developed/built and it's currently in `dev` environment.
This service is updating NOMIS when the domain event above is published.

This means that in `dev` at the moment a duplicated review is recorded in NOMIS (a record is created by our POST request to Prison API and another is created by the sync service).

### Main change
A new feature flag (`feature.review-added-sync-mechanism`, env variable `FEATURE_REVIEWADDEDSYNCMECHANISM`) can be set to determine how a new IEP review is propagated to NOMIS/other services:
- when set to `DOMAIN_EVENT` a domain event will be published
- when set to `PRISON_API` the HTTP request to Prison API will be made

In `dev` environment a domain event will be published for the sync service (and any other services interested in it) while in `preprod`/`prod` environment where the sync service is not deployed to yet an HTTP request to Prison API will be made.

**NOTE**: Either one or the other will be made, as per requirements.

### Other refactorings/changes
- simplified resource class by moving most of the logic into the service class
- separate publishing of domain events from publishing of audit events (the former will be only published depending on the new feature flag but the latter will always be published)
- renamed service methods to make them more explicit and less similar
- added missing tests for `addIepReview()` functionality

### DIFF
--- a/helm_deploy/hmpps-incentives-api/values.yaml
+++ b/helm_deploy/hmpps-incentives-api/values.yaml
@@ -33,6 +33,7 @@ generic-service:
     SPRING_PROFILES_ACTIVE: "logstash"
     APPLICATIONINSIGHTS_CONNECTION_STRING: "InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY)"
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
+    FEATURE_REVIEWADDEDSYNCMECHANISM: "PRISON_API"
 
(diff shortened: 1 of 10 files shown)


neilmendum (approved):
LGTM

PR #198, opened 2022-10-12
INC-823: when SyncPostRequest.current or SyncPatchRequest.current is true then update IEP level with current flag to false

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-823

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonerIepLevelRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonerIepLevelRepository.kt
@@ -10,6 +10,5 @@ interface PrisonerIepLevelRepository : CoroutineCrudRepository<PrisonerIepLevel,
   fun findAllByPrisonerNumberOrderByReviewTimeDesc(prisonerNumber: String): Flow<PrisonerIepLevel>
   fun findAllByBookingIdOrderByReviewTimeDesc(bookingId: Long): Flow<PrisonerIepLevel>
   fun findAllByBookingIdInAndCurrentIsTrueOrderByReviewTimeDesc(bookingIds: List<Long>): Flow<PrisonerIepLevel>
-  suspend fun findFirstByBookingIdAndCurrentIsTrueOrderByReviewTimeDesc(bookingId: Long): PrisonerIepLevel?
   suspend fun findFirstByBookingIdOrderByReviewTimeDesc(bookingId: Long): PrisonerIepLevel?
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
@@ -90,6 +90,10 @@ class PrisonerIepLevelReviewService(
       locationInfo = prisonApiService.getLocationById(prisonerInfo.assignedLivingUnitId, true)
     }
 
+    if (syncPostRequest.current) {
+      updateIepLevelWithCurrentFlagToFalse(bookingId)
+    }
+
     return prisonerIepLevelRepository.save(
       PrisonerIepLevel(
         iepCode = syncPostRequest.iepLevel,
@@ -129,6 +133,10 @@ class PrisonerIepLevelReviewService(
     if (prisonerIepLevel.bookingId != bookingId) {
       log.warn("Patch of PrisonerIepLevel with ID $id failed because provided bookingID ($bookingId) didn't match bookingId on DB record (${prisonerIepLevel.bookingId})")
       throw NoDataFoundException(bookingId)
+    }
+
+    syncPatchRequest.current ?.let {
+      updateIepLevelWithCurrentFlagToFalse(bookingId)
     }
 
     prisonerIepLevel = prisonerIepLevel.copy(
@@ -285,10 +293,7 @@ class PrisonerIepLevelReviewService(
     reviewTime: LocalDateTime,
     reviewerUserName: String,
   ): PrisonerIepLevel {
-    prisonerIepLevelRepository.findFirstByBookingIdAndCurrentIsTrueOrderByReviewTimeDesc(prisonerInfo.bookingId)
-      ?.let {
-        prisonerIepLevelRepository.save(it.copy(current = false))
-      }
+    updateIepLevelWithCurrentFlagToFalse(prisonerInfo.bookingId)
 
     return prisonerIepLevelRepository.save(
       PrisonerIepLevel(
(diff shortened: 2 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -333,6 +338,13 @@ class PrisonerIepLevelReviewService(
     }
   }
 
+  private suspend fun updateIepLevelWithCurrentFlagToFalse(bookingId: Long) {
+    prisonerIepLevelRepository.findFirstByBookingIdAndCurrentIsTrueOrderByReviewTimeDesc(bookingId)
```
xoen, line 342: Bear in mind at the moment (and potentially in the future unless this is enforced at DB level) more than one record for a given prisoner/`bookingId` could have `current=true`.

So maybe this should really get the list of rows with `bookind=$bookingId AND current=true` and loop through them to set current to false on all the rows.

xoen: (so basically replace `.findFirstByBookingIdAndCurrentIsTrue...()` with something like `.findAllByBookingIdAndCurrentIsTrue...()` and the `?.let {}` with some sort of loop that set the current field to `false` on all of them...not the most elegant I know)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonerIepLevelRepository.kt
```diff
@@ -10,6 +10,5 @@ interface PrisonerIepLevelRepository : CoroutineCrudRepository<PrisonerIepLevel,
   fun findAllByPrisonerNumberOrderByReviewTimeDesc(prisonerNumber: String): Flow<PrisonerIepLevel>
   fun findAllByBookingIdOrderByReviewTimeDesc(bookingId: Long): Flow<PrisonerIepLevel>
   fun findAllByBookingIdInAndCurrentIsTrueOrderByReviewTimeDesc(bookingIds: List<Long>): Flow<PrisonerIepLevel>
-  suspend fun findFirstByBookingIdAndCurrentIsTrueOrderByReviewTimeDesc(bookingId: Long): PrisonerIepLevel?
```
neilmendum, line 13: Thought about refactoring this one to drop `First` so it returns `Flow<PrisonerIepLevel>` but decided to remove it and instead use `findAllByBookingIdInAndCurrentIsTrueOrderByReviewTimeDesc` so the repo has fewer methods. Happy to discuss if folks would prefer a different approach

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
neilmendum: Great spot - addressed

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/integration/service/PrisonOffenderEventListenerIntTest.kt
```diff
@@ -92,7 +92,7 @@ internal class PrisonOffenderEventListenerIntTest : SqsIntegrationTestBase() {
   @Test
   fun `prisoner with MERGE numbers is processed`(): Unit = runBlocking {
     // Given
-    val bookingId = 1294134L
+    val bookingId = 1294133L
```
neilmendum, line 95: Despite all integration tests having `repository.deleteAll()` in `BeforeEach` and `AfterEach` function this test was still having issues, so changed the `bookingId` to avoid it

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonerIepLevelRepository.kt
xoen: It's fine, technically in this case you don't need to sort by review time but that's fine :)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -338,9 +338,9 @@ class PrisonerIepLevelReviewService(
     }
   }
 
-  private suspend fun updateIepLevelWithCurrentFlagToFalse(bookingId: Long) {
-    prisonerIepLevelRepository.findFirstByBookingIdAndCurrentIsTrueOrderByReviewTimeDesc(bookingId)
-      ?.let {
+  private suspend fun updateIepLevelsWithCurrentFlagToFalse(bookingId: Long) {
+    prisonerIepLevelRepository.findAllByBookingIdInAndCurrentIsTrueOrderByReviewTimeDesc(listOf(bookingId))
+      .collect {
         prisonerIepLevelRepository.save(it.copy(current = false))
```
xoen, line 344: Just out of curiosity, there is no update method? Is that why we use `copy()`? (When I tried to look up the tree I couldn't find one)

src/main/resources/db/migration/V1_14__prisoner_iep_current_true_constraint.sql
```diff
@@ -0,0 +1 @@
+CREATE UNIQUE INDEX one_booking_id_with_current_true_idx ON prisoner_iep_level (booking_id) WHERE current = true;
```
xoen, line 1: Nice. Would this work on the table which currently probably has some rows violating this? (don't know if it would fail to create the index/migrate or it would instead create this fine)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
neilmendum: Indeed - see `CoroutineCrudRepository`

src/main/resources/db/migration/V1_14__prisoner_iep_current_true_constraint.sql
neilmendum: That's a good point - I'll start up a local Postgres instance and try running the migration against it to see what happens

ushkarev: hmm very good point. maybe we need a cleanup step above?

neilmendum: Indeed - attempting to run that migration against a DB with existing rows violating this results in an error
```
Caused by: org.flywaydb.core.internal.sqlscript.FlywaySqlScriptException: Migration V1_15__prisoner_iep_current_true_constraint.sql failed
----------------------------------------------------------------
SQL State  : 23505
Error Code : 0
Message    : ERROR: could not create unique index "one_booking_id_with_current_true_idx"
  Detail: Key (booking_id)=(1234567) is duplicated.
```
Will

ushkarev (approved):

PR #200, opened 2022-10-13
INC-675: eclipse-temurin docker image

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-675

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:18-slim AS builder
+FROM eclipse-temurin:18-jre-jammy AS builder
 
 ARG BUILD_NUMBER
 ENV BUILD_NUMBER ${BUILD_NUMBER:-1_0_0}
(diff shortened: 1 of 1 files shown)


ushkarev (approved):
i'd have expected the builder to use a JDK-containing variant like `eclipse-temurin:18` or `eclipse-temurin:18-jdk` but perhaps gradle downloads some random jdk as part of it's build steps?

neilmendum:
@ushkarev indeed - I would've thought the same

Looking more into `./gradlew assemble` a [few lines below](https://github.com/ministryofjustice/hmpps-incentives-api/blob/9f603f9a53360b2ba245ddef1ba32f327327b4b0/Dockerfile#L8)

Did a bit of Googling about _Gradle JDK_ found [this answer](https://discuss.gradle.org/t/how-does-gradle-determine-which-jdk-to-use/19896/2). I'm assuming it picks up the JDK executor from [hmpps-circleci-orb](https://github.com/ministryofjustice/hmpps-circleci-orb) and uses that to compile

PR #203, opened 2022-10-17
INC-802: Get prison IEP levels from Prison API (not incentives DB)

xoen:
The incentives-api database currently holds a list of IEP levels for the prisons and their corresponding default level.

Unfortunately that list is not kept in sync which means is not reflecting the real list of IEP levels for some prisons and/or their default level.

As this is not synchronised with NOMIS we're removing this DB table and replacing it with a call to Prison API's `GET /api/agencies/$prisonId/iepLevels`.

**NOTE**: This endpoint wasn't exposing the default IEP level but [this was added recently](https://github.com/ministryofjustice/prison-api/pull/1421), check this is deployed to `prod`.

### Other changes
- removed FK constraint in `prisoner_iep_level` table to `iep_prison.iep_code` (as this DB table is going)
- moved Prison API-related dto classes under `dto.prisonapi` package for clarity
- removed now unused services in `IepLevelResource`


neilmendum (approved):
LGTM

PR #207, opened 2022-10-20
INC-872: sync DELETE endpoint

xoen:
`DELETE /iep/sync/booking/{bookingId}/id/{id}`

- delete an IEP review by id
- publishes a `IEP_REVIEW_DELETED` domain/audit events
- when the deleted IEP review had `current=true`, sets another one (latest) as current so that there is one current IEP review for that `bookingId`


neilmendum (approved):
LGTM

PR #212, opened 2022-10-21
INC-877 remove iep_level table

Mjwillis:
Signed-off-by: mjwillis <michael.willis@digital.justice.gov.uk>


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonApiService.kt
```diff
@@ -106,3 +127,11 @@ class PrisonApiService(
       .awaitBody()
   }
 }
+
+data class IncentiveLevel(
```
neilmendum, line 131: Could move this to `prisonapidto.kt` which is where others are defined?

PR #213, opened 2022-10-21
INC 878 Provide the feature switch for incentives DB/NOMIS

Mjwillis:
- INC-877 remove iep_level table
- INC-878 Switch feature


Mjwillis:
Currently 4 tests failing because expecting to read from Incentives DB

helm_deploy/values-dev.yaml
```diff
@@ -13,6 +13,7 @@ generic-service:
     API_BASE_URL_OAUTH: https://sign-in-dev.hmpps.service.justice.gov.uk/auth
     API_BASE_URL_PRISON: https://api-dev.prison.service.justice.gov.uk
     FEATURE_REVIEWADDEDSYNCMECHANISM: "DOMAIN_EVENT"
+    FEATURE_INCENTIVES_DATA_SOURCE_OF_TRUTH: false
```
xoen, line 16: I'm new to Spring Boot but you may need to add it to `src/resources/(application.yml|application-dev.yml)` (similar to what done [for the other property](https://github.com/ministryofjustice/hmpps-incentives-api/blob/main/src/main/resources/application.yml#L5-L12))

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/HmppsIncentivesApiExceptionHandler.kt
```diff
@@ -109,6 +110,20 @@ class HmppsIncentivesApiExceptionHandler {
       )
   }
 
+  @ExceptionHandler(IncentiveReviewNotFoundException::class)
+  fun handleRoleNotFoundException(e: IncentiveReviewNotFoundException): ResponseEntity<ErrorResponse?>? {
```
xoen, line 114: probably the method name should be adjusted to something like `handleIncentiveReviewNotFoundException()` :)

I wonder if a method with a more generic `Exception` type could handle all these exceptions that we want to result in a `404 Not Found`, e.g. maybe it's possible to "stack up" several different `@ExceptionHandler` annotations?

e.g.

```Kotlin
@ExceptionHandler(NoPrisonersAtLocationException::class)
@ExceptionHandler(IncentiveReviewNotFoundException::class) 
// etc...
fun handleNotFoundExceptions(e: Exception): ResponseEntity<ErrorResponse?>? {
    log.debug("Not found exception caught: {}", e.message)
    return ResponseEntity
      .status(NOT_FOUND)
      .body(
        ErrorResponse(
          status = NOT_FOUND,
          userMessage = "Not Found: ${e.message}",
          developerMessage = e.message
        )
      )
}
```

although the log message would be less specific. I'm not even sure if that is possible, just a thought, this is working and fine :)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelResource.kt
```diff
@@ -100,12 +100,10 @@ class IepLevelResource(
   suspend fun getPrisonerIepLevelHistory(
     @Schema(description = "Booking Id", example = "3000002", required = true, type = "integer", format = "int64", pattern = "^[0-9]{1,20}$")
     @PathVariable bookingId: Long,
-    @Schema(description = "Use NOMIS data", example = "true", required = false, defaultValue = "true", hidden = true, type = "boolean", pattern = "^[true|false]$")
-    @RequestParam(defaultValue = "true", value = "use-nomis-data", required = false) useNomisData: Boolean = true,
```
xoen, line 104: As mentioned this would be an API breaking change but since no consumers seem to explicitly pass/use this param (e.g. they default to getting the data from NOMIS) we think that's OK.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/FeatureFlagsService.kt
```diff
@@ -12,9 +12,15 @@ enum class ReviewAddedSyncMechanism {
 class FeatureFlagsService(
   @Value("\${feature.review-added-sync-mechanism}")
   val reviewAddedSyncMechanism: String,
+  @Value("\${feature.incentives-data-source-of-truth:false}")
+  val incentivesDataSourceOfTruth: Boolean
```
neilmendum, line 16: You could name this something like `incentivesDbSourceOfTruth` to be a bit more explicit?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -50,24 +52,27 @@ class PrisonerIepLevelReviewService(
...
     withDetails: Boolean = true,
     useClientCredentials: Boolean = false,
   ): IepSummary {
-    return if (useNomisData) {
+    return if (!featureFlagsService.isIncentivesDataSourceOfTruth()) {
       prisonApiService.getIEPSummaryForPrisoner(bookingId, withDetails, useClientCredentials)
     } else {
-      buildIepSummary(prisonerIepLevelRepository.findAllByBookingIdOrderByReviewTimeDesc(bookingId), getIncentiveLevels(), withDetails)
+      val reviews = prisonerIepLevelRepository.findAllByBookingIdOrderByReviewTimeDesc(bookingId)
+      if (reviews.count() == 0) throw IncentiveReviewNotFoundException("No Incentive Reviews for booking ID $bookingId")
```
xoen, line 62: Nice, I like the throwing of an exception when we detect a problematic state 👍 

Can I suggest we also log a debug message on top of this (can help debug these)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -249,27 +265,43 @@ class PrisonerIepLevelReviewService(
 
   private suspend fun getIepLevelForReviewType(prisonerInfo: PrisonerAtLocation, reviewType: ReviewType): String {
     val iepLevelsForPrison = iepLevelService.getIepLevelsForPrison(prisonerInfo.agencyId, useClientCredentials = true)
+    val defaultLevel = iepLevelsForPrison.find(IepLevel::default) ?: iepLevelsForPrison.first() // if no default use the lowest sequence
```
xoen, line 268: Is it "normal" for prisons to not have a default IEP level?

If they should always have a default IEP level I think it's worth keeping [that exception that is now removed](https://github.com/ministryofjustice/hmpps-incentives-api/pull/213/files#diff-8bf11d470f9e689ef0ce56f887d92fb6c5c77d56adf2a909d15d4a8d00096a64L253) as it highlights a data problem with an individual prison.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
xoen: Actually the exception is probably enough as we see that in AppInsights anyway so don't worry :)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -249,27 +265,43 @@ class PrisonerIepLevelReviewService(
...
     val iepLevel = when (reviewType) {
-      ReviewType.INITIAL -> iepLevelsForPrison.find(IepLevel::default) ?: throw Exception("No default IEP level found for Prison '${prisonerInfo.agencyId}'")
-
+      ReviewType.INITIAL -> {
+        defaultLevel // admission should always be the default
```
neilmendum, line 271: Looks like we're changing the logic here - wonder if we want to create a new story to cover it and make the change in a separate PR?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/FeatureFlagsService.kt
Mjwillis: Do you have any ideas have you to change or could do so in the next PR?

Mjwillis:
@neilmendum are you happy to approve today or do you need further changes? I'm happy to pick up additional tweaks on Wednesday but I don't think we should leave his PR outstanding for a number of days. It's your call however

helm_deploy/values-dev.yaml
xoen: Sorry just noticed this is added to the `application.yml` below 👌

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelResourceTest.kt
```diff
@@ -96,51 +92,6 @@ class IepLevelResourceTest : SqsIntegrationTestBase() {
       )
   }
 
-  @Test
```
xoen, line 99: What happened to this test and the other to test these two endpoints?

These endpoints are still there AFAICT
- `GET /iep/reviews/booking/{bookingId}`
- `POST /iep/reviews/bookings`

xoen: Oh the tests are simply moved to `IepLevelResourceForNomisDataTest.kt` 👌

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
Mjwillis: But Nomis doesn't do that - so will be out of sync

neilmendum:
@Mjwillis replying to https://github.com/ministryofjustice/hmpps-incentives-api/pull/213#issuecomment-1288899704

My concern with this PR is that it seems to me like it's doing 2 things

1. Adding a feature switch
2. Changing the logic when processing `ADMISSION` and `TRANSFERRED` domain events

I see your point on https://github.com/ministryofjustice/hmpps-incentives-api/pull/213#discussion_r1003206924 that for `TRANSFERRED` we will source `iepHistory` from the Incentives DB with the call to `PrisonerIepLevelReviewService.getPrisonerIepLevelHistory` if the feature flag is `true`. Then we're being defensive against data issues by catching `IncentiveReviewNotFoundException`

However in both cases we're now changing the logic to now fallback to `defaultLevel`, rather than throwing an exception and putting the message back on the queue to be reprocessed. It's this which I think should be discussed with Steve H and we can raise a new story to update that specific logic

Mjwillis:
We really cannot do have it going on to the DLQ - Nomis can handle the "BAD" data and so must we or we will be out of sync with NOMIS.  We cannot be out of sync with NOMIS in any circumstance which will happen if we exception here.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/FeatureFlagsService.kt
Mjwillis: updated

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/HmppsIncentivesApiExceptionHandler.kt
Mjwillis: updated

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelResource.kt
Mjwillis: correct

Mjwillis:
I've made a number of changes as requested apart from the transfer logic which we really have to fix.  I really appreciate if could you review again.

ushkarev:
no complaints from me other than a tiny question https://github.com/ministryofjustice/hmpps-incentives-api/pull/213#discussion_r1004515817 :)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
Mjwillis: Yes

PR #216, opened 2022-10-25
INC-846 & INC-842: Return basic information about prisoners in a given location

ushkarev:
…which will be iteratively enhanced to show information relevant to incentive reviews in future.

PR #227, opened 2022-10-27
INC-816: Wire up offender search service to return a single offender by prisoner number

ushkarev:
…which will be needed for future "date of next review" calculations.

Also, changes to use system token to access offender search: Essentially, we only require the `MAINTAIN_IEP` role, and this makes sense for the data we actually return to the caller. But to calculate next review dates, we need to load more information on the given offender and the calling user may not have the roles to do so.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/OffenderSearchService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/OffenderSearchService.kt
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.incentivesapi.service
 import org.springframework.stereotype.Service
 import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.reactive.function.client.awaitBody
+import uk.gov.justice.digital.hmpps.incentivesapi.dto.OffenderSearchPrisoner
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.OffenderSearchPrisonerList
 
 @Service
@@ -25,4 +26,14 @@ class OffenderSearchService(private val offenderSearchWebClient: WebClient) {
       )
       .retrieve()
       .awaitBody<OffenderSearchPrisonerList>()
+
+  /**
+   * Gets one offender record
+   * Requires ROLE_PRISONER_SEARCH or ROLE_VIEW_PRISONER_DATA role
+   */
+  suspend fun getOffender(prisonerNumber: String) =
+    offenderSearchWebClient.get()
+      .uri("/prisoner/$prisonerNumber")
+      .retrieve()
+      .awaitBody<OffenderSearchPrisoner>()
 }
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/OffenderSearchService.kt
```diff
@@ -24,5 +25,15 @@ class OffenderSearchService(private val offenderSearchWebClient: WebClient) {
...
       )
       .retrieve()
-      .awaitBody<OffenderSearchPrisonerResponse>()
+      .awaitBody<OffenderSearchPrisonerList>()
+
+  /**
+   * Gets one offender record
+   * Requires ROLE_PRISONER_SEARCH or ROLE_VIEW_PRISONER_DATA role
+   */
+  suspend fun getOffender(prisonerNumber: String) =
```
xoen, line 34: Nice 👍 

As this and the other method needs "special" roles and we probably call these endpoints internally to enrich the response with data the consumer didn't directly ask...we may want these to happen using the `incentives-api` client.

e.g. update the `offenderSearchWebClient()` bean [to do something similar to the `prisonWebClientClientCredentials()` bean](https://github.com/ministryofjustice/hmpps-incentives-api/blob/5815d74e87083abbb614d9b9e1b2b7dd8fd5c712/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/WebClientConfiguration.kt#L44).

I have reached the conclusion that's the safest thing if we start to make these requests to Offender Search API, as otherwise we may accidentally break some of the API consumers.

neilmendum: We'll want to add these roles to the `hmpps-incentives-api` client credentials? I can revert changes made with https://github.com/ministryofjustice/hmpps-incentives-api/pull/216#discussion_r1005415048

ushkarev: yeah we should _if_ you agree that this switch is a good idea @neilmendum. trouble is, i don't know what roles typical users have nor what our target audience is likely to. maybe it's not something we need to worry about cos everybody has the search role? on the other hand, we only _need_ to require the relevant (MAINTAIN_IEP) role so should we also pile on external data's dependent roles… i feel that we shouldn't.

neilmendum: > we only need to require the relevant (MAINTAIN_IEP) role

Agreed. This seems like a good approach to me and we shouldn't assume callers have the search role. Us using Offender Search is an implementation detail, which I don't think callers should need to worry about, and this solves that issue

xoen (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/OffenderSearchService.kt
xoen: Yeah I agree with both of you and use the system credentials for this "implementation detail" requests to Offender Search API is the safer approach 👌 

We only need to be sure our client has these roles. Otherwise we'd risk to accidentally break consumers that were previously using the API without problems (e.g. especially when we'll add these requests to get ACCT status to determine next review date)

PR #230, opened 2022-10-27
Have renovate-bot preserve existing semver ranges

ushkarev:
…so that it does not suggest too many unnecessary updates to patch-level versions.

E.g. the HMPPS CircleCI orb is fine being pinned at the minor level because CircleCI will always use the latest patch version then anyway

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,7 +1,8 @@
 {
   "extends": [
     "config:base",
-    ":disableDependencyDashboard"
+    ":disableDependencyDashboard",
+    ":preserveSemverRanges"
   ],
   "assigneesFromCodeOwners": true,
   "timezone": "Europe/London",


neilmendum (approved):
Thanks for looking at this - it's also been on my mind because it would be nice for Renovate to ignore patch versions for the HMPPS Orb

When trialing Renovate I spent [quite a bit of time](https://github.com/ministryofjustice/renovate-poc/commits/main) trying to get it to work, but didn't manage it

Good shout on `preserveSemverRanges` - let's see how that goes!

ushkarev:
i just noticed that incentives-ui had this flag set but this repo didn't. i recall you'd fought renovate quite a lot to get it to behave; hope i'm not redoing something you'd already tried ;)

PR #231, opened 2022-10-28
INC-874: Find best level for transfers when the same is not available

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
```diff
@@ -16,4 +17,56 @@ class IepLevelService(
...
+      .map {
+        KnownLevel(
+          code = it.iepLevel,
+          available = it.active && levelCodesAvailableInPrison.contains(it.iepLevel),
+        )
+      }
+
+    val indexOfTargetLevelCode = allKnownLevels.indexOfFirst { it.code == levelCode }
+    if (indexOfTargetLevelCode == -1) {
+      // target level does not appear in known levels so can only fall back to prison’s default level
```
ushkarev, line 54: extreme edge case where the level before is not available globally (this should not be possible)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
```diff
@@ -16,4 +17,56 @@ class IepLevelService(
...
+      }
+    }
+    // find the first available level lower or same as target level
+    for (someLevel in allKnownLevels.withIndex().reversed()) {
+      if (someLevel.index <= indexOfTargetLevelCode && someLevel.value.available) {
+        return someLevel.value.code
+      }
+    }
+    // no levels were available so can only fall back to prison’s default level
+    return defaultLevelCode
```
ushkarev, line 70: this should be unreachable

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
```diff
@@ -16,4 +17,56 @@ class IepLevelService(
...
+        if (it.default) {
+          maybeDefaultLevelCode = it.iepLevel
+        }
+        it.iepLevel
+      }
+    val defaultLevelCode = maybeDefaultLevelCode
+      // NOMIS traditionally takes the first if no default was found
+      ?: levelCodesAvailableInPrison.firstOrNull()
+      // there are no available incentive levels at all at this prison
+      ?: throw DataIntegrityException("$prisonId has no available incentive levels")
```
ushkarev, line 40: serious data integrity problem: this prison has no incentive levels. you probably can't even assign a level to a person

Mjwillis: agreed

Mjwillis (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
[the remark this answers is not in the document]
ushkarev: this is the extreme edge case where a prison has no incentive levels at all. we cannot choose a level when literally none are available. i doubt one could set an IEP level in NOMIS for a prisoner at such an establishment.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -301,19 +307,17 @@ class PrisonerIepLevelReviewService(
               withDetails = true,
               useClientCredentials = true
             ).iepDetails
-          val iepLevelBeforeTransfer =
```
ushkarev, line 304: there was a type mismatch in the calculation of `iepLevelBeforeTransfer`

`iepHistory.sortedBy(IepDetail::iepTime).lastOrNull { it.agencyId != prisonerInfo.agencyId }?.iepCode` is `String?`
`defaultLevel` is `IepLevel`
making `iepLevelBeforeTransfer` an `Any`

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
```diff
@@ -16,4 +17,56 @@ class IepLevelService(
...
   }
+
+  /**
+   * Because not all levels are enabled at all prisons, and we want to maintain a
+   * person’s level when they are transferred, we need to find the "nearest highest" level
+   * according to HMPPS policy. Falls back to prison's default level when there's nothing else to do.
+   */
+  suspend fun findNearestHighestLevel(prisonId: String, levelCode: String): String {
+    var maybeDefaultLevelCode: String? = null
+    val levelCodesAvailableInPrison = getIepLevelsForPrison(prisonId, true)
```
ushkarev, line 28: i don't know whether to `useClientCredentials` here :/

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewServiceTest.kt
```diff
@@ -387,9 +377,9 @@ class PrisonerIepLevelReviewServiceTest {
       // When
       prisonerIepLevelReviewService.processOffenderEvent(prisonOffenderEvent)
 
-      // Then - the prisoner was on STD at BXI so they on this level
+      // Then - the prisoner was on BAS at BXI so they should remain on this level
       val expectedPrisonerIepLevel = PrisonerIepLevel(
-        iepCode = "STD",
+        iepCode = "BAS",
```
ushkarev, line 382: switched to testing using Basic level because it's not the default in this prison meaning this test actually shows the level being preserved and definitely not resetting to a default

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
ushkarev: also i can't figure out the role needed to call `prisonApiService.getIepLevels()` / GET "/api/reference-domains/domains/{domain}/codes/{code}". possibly none??

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -301,19 +307,17 @@ class PrisonerIepLevelReviewService(
...
+          val levelCodeBeforeTransfer =
             iepHistory.sortedBy(IepDetail::iepTime).lastOrNull { it.agencyId != prisonerInfo.agencyId }?.iepCode
-              ?: defaultLevel // if no previous prison
-          iepLevelsForPrison.find { it.iepLevel == iepLevelBeforeTransfer }
-            ?: defaultLevel // if we don't match a level - go to the default.
+              ?: defaultLevelCode // if no previous prison
+          iepLevelService.findNearestHighestLevel(prisonerInfo.agencyId, levelCodeBeforeTransfer)
```
neilmendum, line 313: We could pass `iepLevelsForPrison` into this function to avoid needing to duplicate the lookup?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
neilmendum: > i don't know whether to useClientCredentials here or not :/

We need to because the request originates from `processOffenderEvent` which is invoked when we receive an SQS message, which unlike HTTP endpoints being called by other APIs, does not have client credentials in the header so we need to use our own when making subsequent HTTP calls

And yep looking at https://github.com/ministryofjustice/prison-api/blob/main/src/main/java/uk/gov/justice/hmpps/prison/api/resource/ReferenceDomainResource.java - looks like no role is required for that endpoint

neilmendum (commented):
Looks good! added 1 minor comment

neilmendum (approved):
LGTM

PR #235, opened 2022-10-31
INC-890: Hard-code dates within incentive level date calculation tests

ushkarev:
…making them easier to read now that the "now" time is fixed


neilmendum (approved):
LGTM

PR #236, opened 2022-11-01
INC-852: Take open ACCT status when determining next review date

xoen:
The rule to determine the next review date _for prisoners on 'Basic'_ is slightly different for prisoners at risk of self harm (have an ACCT alert active and not expired, aka is a prisoner "with open ACCT") their next review date is earlier.

When someone was on 'Basic' and they couldn't move up to an higher level (e.g. they were confirmed on 'Basic' level) the next review would normally be after 28 days. However for prisoners with an open ACCT in this case it would be after 14 days instead.

> PF 5.16 _Prisoners placed on Basic_ must be reviewed within 7 days and if not suitable to return to Standard level further reviews must be undertaken at least every 28 days thereafter
>
> _Exception for those identified as at Risk of Suicide and Self-harm (ACCT)_ and for young people, where further reviews must be undertaken at least every 14 days thereafter

(emphasis on bits relative for this PR/ticket mine)

**NOTE**: ACCT is only affecting prisoners on Basic, not on other levels. This is why in the code this is part of the "basic rules" calculation.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
@@ -38,7 +38,7 @@ class IncentiveSummaryService(
     sortBy: SortColumn = SortColumn.NAME,
     sortDirection: Sort.Direction = Sort.Direction.ASC
   ): BehaviourSummary = coroutineScope {
-    val prisoners = prisonApiService.findPrisonersAtLocation(prisonId, locationId).toList()
+    val prisoners = prisonApiService.findPrisonersAtLocation(locationId).toList()
     if (prisoners.isEmpty()) throw NoPrisonersAtLocationException(prisonId, locationId)
 
     val iepLevelsDeferred = async { iepLevelService.getIepLevelsForPrison(prisonId) }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateService.kt
@@ -53,5 +53,4 @@ class NextReviewDateService(private val input: NextReviewDateInput) {
 
     return isOnBasic() && iepDetails.size >= 2 && iepDetails[1].iepLevel == BASIC
   }
-
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateService.kt
@@ -7,6 +7,7 @@ private const val BASIC = "Basic"
 
 data class NextReviewDateInput(
   val iepDetails: List<IepDetail>,
+  val hasAcctOpen: Boolean,
 )
 
 class NextReviewDateService {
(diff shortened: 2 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonApiService.kt
```diff
@@ -56,7 +56,7 @@ class PrisonApiService(
       }
   }
 
-  suspend fun findPrisonersAtLocation(prisonId: String, locationId: String): Flow<PrisonerAtLocation> =
+  suspend fun findPrisonersAtLocation(locationId: String): Flow<PrisonerAtLocation> =
```
xoen, line 59: `prisonId` was not used in the method, the Prison API endpoint only uses the `locationId` anyway.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
```diff
@@ -38,7 +38,7 @@ class IncentiveSummaryService(
     sortBy: SortColumn = SortColumn.NAME,
     sortDirection: Sort.Direction = Sort.Direction.ASC
   ): BehaviourSummary = coroutineScope {
-    val prisoners = prisonApiService.findPrisonersAtLocation(prisonId, locationId).toList()
+    val prisoners = prisonApiService.findPrisonersAtLocation(locationId).toList()
```
xoen, line 41: `prisonId` was not used, [see other comment](https://github.com/ministryofjustice/hmpps-incentives-api/pull/236/files#r1010289259).

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -45,21 +45,32 @@ class PrisonerIepLevelReviewService(
...
 
-  private fun setNextReviewDate(iepSummary: IepSummary): IepSummary {
+  private suspend fun setNextReviewDate(iepSummary: IepSummary): IepSummary {
     if (iepSummary.iepDetails.isEmpty()) {
       throw IllegalArgumentException("Cannot set nextReviewDate without iepDetails for bookingId = ${iepSummary.bookingId}")
     }
 
-    iepSummary.nextReviewDate = NextReviewDateService().calculate(
+    // NOTE: `iepSummary.prisonerNumber` could be `null` (when data is coming from NOMIS)
+    val prisonerNumber = iepSummary.prisonerNumber ?: run {
```
xoen, line 60: As mentioned in the comment [in the `IepSummary`class the `prisonerNumber` is optional](https://github.com/ministryofjustice/hmpps-incentives-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt#L22) and could be blank.

`prisonerNumber` would be blank when the data is coming from NOMIS/Prison API, but it will be populated when it's coming from our database (as [the column is not NULL-able](https://github.com/ministryofjustice/hmpps-incentives-api/blob/main/src/main/resources/db/migration/V1_7__prisoner_iep_level.sql#L6)).

Once we stop getting reviews from Prison API we can probably make the `prisonerNumber` mandatory and remove this logic to make this additional request to get the `prisonerNumber`.

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewServiceTest.kt
```diff
@@ -354,25 +410,37 @@ class PrisonerIepLevelReviewServiceTest {
...
+      whenever(prisonApiService.getPrisonerInfo(prisonerNumber, true)).thenReturn(prisonerAtLocation)
       whenever(prisonApiService.getLocationById(prisonerAtLocation.assignedLivingUnitId, true)).thenReturn(location)
       whenever(prisonApiService.getIepLevelsForPrison(prisonerAtLocation.agencyId, true)).thenReturn(basicStandardEnhanced)
       whenever(prisonApiService.getIepLevels()).thenReturn(globalIncentiveLevels.asFlow())
       val iepDetails = listOf(
         iepDetail(prisonerAtLocation.agencyId, "Standard", "STD", LocalDateTime.now()),
         iepDetail("BXI", "Basic", "BAS", LocalDateTime.now().minusDays(1)),
       )
-      val iepSummary = iepSummary(iepDetails = iepDetails)
+      val iepSummary = iepSummary("Basic", bookingId, iepDetails)
```
xoen, line 427: You'll see that there are a few tweaks, they're not just to improve readability.

The test was failing and after digging I realised that the wiring of the mocks was returning conflicting responses with each other.

e.g. a request would return `bookingId` X and then a separate request would say `bookingId` is Y

This was because of some test data using default values. I made this more explicit and in general pass the desired `bookingId`/`prisonerNumber`.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateService.kt
```diff
@@ -7,37 +7,50 @@ private const val BASIC = "Basic"
...
+
+  private fun isOnBasic(): Boolean {
+    return lastReview().iepLevel == BASIC
+  }
+
+  private fun lastReviewDate(): LocalDate {
+    return lastReview().iepDate
+  }
+
+  private fun wasConfirmedBasic(): Boolean {
```
neilmendum, line 51: Is _Confirmed_ wording from the policy? If not could name this something like `lastTwoReviewsOnBasic`?

neilmendum (commented):
Looks good! Made 1 minor comment about function naming

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateService.kt
xoen: Policy says "if _not suitable to return_ to Standard level", where "return to Standard" means get to an higher level.

The "not suitable to returns" means they were on 'Basic' but they were reviewed again and confirmed they would stay at the same level.

It's not wording from the policy, but regarding the method name I didn't want to use a name which was too low-level/tied to the last reviews being on basic even tho this is what about, hence the "wasConfirmedBasic" name. If it makes sense.

xoen: e.g. I named it after the concept rather than the actual implementation/logic (or I would have called it `hasAtLeastTwoReviewsAndTheyAreBasicLevel()` or something 😝)

The idea is that if you really are curious about the implementation you can read the method body but when you're going through the calculation logic you only care about the high level concept.

xoen: It's a private method anyway, if policy had a short word for the concept it would be nice to use but otherwise we could almost call this anything :)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
ushkarev: might be worth saving `prisonerNumber` into `iepSummary` in case it's needed in other calls later?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateService.kt
neilmendum: Makes sense 😄

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
xoen: I see your point. Given this fallback logic is gonna disappear when we only return data from the incentives' DB I'm very tempted to follow the single responsibility principle and _not_ set anything else aside from the `nextReviewDate`. 
Unless you have strong feelings about this of course :)

ushkarev (approved):
look alright to me. but you know that i'd have not been able to stop myself using a lazy getter instead of the methods like `private fun isOnBasic(): Boolean`

PR #237, opened 2022-11-02
INC-848: include simplest calculation for date of next review

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-848

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
@@ -5,14 +5,18 @@ import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.toList
 import org.springframework.stereotype.Service
 import org.springframework.web.reactive.function.client.WebClientResponseException.NotFound
+import uk.gov.justice.digital.hmpps.incentivesapi.config.ListOfDataNotFoundException
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.IncentiveReview
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.IncentiveReviewResponse
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.prisonapi.CaseNoteUsage
+import uk.gov.justice.digital.hmpps.incentivesapi.jpa.repository.PrisonerIepLevelRepository
+import java.time.LocalDate
 
 @Service
 class IncentiveReviewsService(
   private val offenderSearchService: OffenderSearchService,
   private val prisonApiService: PrisonApiService,
+  private val prisonerIepLevelRepository: PrisonerIepLevelRepository,
 ) {
   /**
    * Returns incentive review information for a given location within a prison
@@ -39,6 +43,14 @@ class IncentiveReviewsService(
     val deferredPositiveCaseNotesInLast3Months = async { getCaseNoteUsage("POS", "IEP_ENC", prisonerNumbers) }
     val deferredNegativeCaseNotesInLast3Months = async { getCaseNoteUsage("NEG", "IEP_WARN", prisonerNumbers) }
 
+    val bookingIds = offenders.content.map { p -> p.bookingId.toLong() }.toSet()
+    val nextReviewDates = getNextReviewDatesForOffenders(bookingIds)
+
+    val bookingIdsMissingIepReviews = bookingIds subtract nextReviewDates.keys
+    if (bookingIdsMissingIepReviews.isNotEmpty()) {
+      throw ListOfDataNotFoundException(bookingIdsMissingIepReviews)
+    }
+
     val positiveCaseNotesInLast3Months = deferredPositiveCaseNotesInLast3Months.await()
     val negativeCaseNotesInLast3Months = deferredNegativeCaseNotesInLast3Months.await()
 
@@ -56,8 +68,9 @@ class IncentiveReviewsService(
           positiveBehaviours = positiveCaseNotesInLast3Months[it.prisonerNumber]?.totalCaseNotes ?: 0,
           negativeBehaviours = negativeCaseNotesInLast3Months[it.prisonerNumber]?.totalCaseNotes ?: 0,
           acctOpenStatus = it.acctOpen,
+          nextReviewDate = nextReviewDates[it.bookingId.toLong()]!!.nextReviewDate,
         )
-      }
+      }.sortedBy { it.nextReviewDate }
     )
   }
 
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
```diff
@@ -56,8 +68,9 @@ class IncentiveReviewsService(
           positiveBehaviours = positiveCaseNotesInLast3Months[it.prisonerNumber]?.totalCaseNotes ?: 0,
           negativeBehaviours = negativeCaseNotesInLast3Months[it.prisonerNumber]?.totalCaseNotes ?: 0,
           acctOpenStatus = it.acctOpen,
+          nextReviewDate = nextReviewDates[it.bookingId.toLong()]!!.nextReviewDate,
```
xoen, line 71: isn't the `!!` going to throw a null pointer exception? I wouldn't think the value for any `bookingId` shouldn't be missing or we would get a `ListOfDataNotFoundException` above right?

Is it because `[]` returns `T?` and Kotlin is complaining unless you use `!!`? Maybe it doesn't matter in this case as we know we should never get a map with some of the `bookingId` missing right?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
```diff
@@ -76,4 +89,21 @@ class IncentiveReviewsService(
...
+  private suspend fun getNextReviewDatesForOffenders(bookingIds: Set<Long>): Map<Long, IepResult> {
+    return prisonerIepLevelRepository.findAllByBookingIdInAndCurrentIsTrueOrderByReviewTimeDesc(bookingIds.toList())
+      .toList()
+      .groupBy { it.bookingId }
+      .map {
+        IepResult(
+          bookingId = it.key,
+          nextReviewDate = it.value.first().reviewTime.plusYears(1).toLocalDate()
+        )
+      }.associateBy(IepResult::bookingId)
```
xoen, line 102: So we'll replace this with a call to `NextReviewDateService` at some point 👌

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
```diff
@@ -76,4 +89,21 @@ class IncentiveReviewsService(
 
   private fun calcTypeCount(caseNoteUsage: List<CaseNoteUsage>): Int =
     caseNoteUsage.map { it.numCaseNotes }.fold(0) { acc, next -> acc + next }
+
+  private suspend fun getNextReviewDatesForOffenders(bookingIds: Set<Long>): Map<Long, IepResult> {
```
xoen, line 93: This method will probably need to get or receive the list of Offender Search prisoners for things like the DoB, ACCT, etc

xoen (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
neilmendum: > Is it because [] returns T? and Kotlin is complaining unless you use !!? Maybe it doesn't matter in this case as we know we should never get a map with some of the bookingId missing right?

Exactly, on both points. I'm equally not a huge fan of `!!` but we've protected against this handling `ListOfDataNotFoundException`

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
ushkarev, line 93: given a person can't appear more than once in offender search results, there's probably no need to convert `bookingIds` from list to set back to list

neilmendum: Yes. It's probably worth mentioning that I haven't included a branch for the `isIncentivesDataSourceOfTruth` and we're _only_ getting data from the Incentives DB because I've assumed that migration will have taken place by the time this screen is deployed

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveReviewsResourceTest.kt
```diff
@@ -1,15 +1,61 @@
...
+      prisonerNumber = "A1409AE",
+    )
+
+    persistPrisonerIepLevel(
+      bookingId = 110002,
+      prisonerNumber = "G6123VU",
+    )
+  }
+
+  private val iepTime: LocalDateTime = LocalDateTime.now()
```
ushkarev, line 36: doesn't matter hugely (or now), but would probs be nice to force the system clock "now" during these tests so that one could see all date asserts explicitly as dates.

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsServiceTest.kt
```diff
@@ -198,11 +210,110 @@ class IncentiveReviewsServiceTest {
...
+  @Test
+  fun `oldest date of next review is first`(): Unit = runBlocking {
+    // Given
+    val prisonerNumber = "G6123VU"
+    whenever(prisonApiService.getLocation(any())).thenReturnLocation("MDI-2-1")
+    whenever(offenderSearchService.findOffenders(any(), any(), any(), any()))
+      .thenReturnOffenders(
+        listOf(
+          offenderSearchPrisoner(prisonerNumber, "110001"),
+          offenderSearchPrisoner(prisonerNumber, "110002")
```
ushkarev, line 228: minor thing, but it'd make sense to have different prisoner numbers as well as different booking ids. i know that won't affect the calculations of what these tests are actually asserting, but the tests would read more like reality

ushkarev (approved):
only 1 minor comment about some tests being more realistic; your call

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
xoen: Yeah good point, I noticed that and it makes sense as new table/endpoint is assumed to only work off incentives DB data 👌 (and it's nice and simpler without "weird" branches :))

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveReviewsResourceTest.kt
xoen: IMHO it doesn't matter when the time is not used or used only to check relative time differences like in this case (the assertion is on the +1 year difference)

In fact I personally find `.now()`/`.now().plusSomething()` much easier on the eye than hardcoding all the dates (which is manual and error prone, e.g. copy and paste and it's possible to end up with an older review with a date/time which is not quite right).

Just my 2 eurocents :P

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
neilmendum: Good point - addressed https://github.com/ministryofjustice/hmpps-incentives-api/pull/237/commits/0f7365b19d018a307509345d6d2bcbfacfa0485e

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsServiceTest.kt
neilmendum: Good point - addressed https://github.com/ministryofjustice/hmpps-incentives-api/pull/237/commits/0f7365b19d018a307509345d6d2bcbfacfa0485e

ushkarev (approved):

PR #238, opened 2022-11-02
INC-838: total of Overdue Prisoner Review

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-838

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
@@ -10,6 +10,7 @@ import uk.gov.justice.digital.hmpps.incentivesapi.dto.IncentiveReview
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.IncentiveReviewResponse
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.prisonapi.CaseNoteUsage
 import uk.gov.justice.digital.hmpps.incentivesapi.jpa.repository.PrisonerIepLevelRepository
+import java.time.Clock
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
```diff
@@ -50,6 +52,7 @@ class IncentiveReviewsService(
     if (bookingIdsMissingIepReviews.isNotEmpty()) {
       throw ListOfDataNotFoundException(bookingIdsMissingIepReviews)
     }
+    val overdueCount = nextReviewDates.map { r -> r.value.nextReviewDate }.count { it.isBefore(LocalDate.now(clock)) }
```
ushkarev, line 55: this only counts the number overdue on this page

neilmendum: D'oh! I see what you mean as we're querying `offenderSearchService.findOffenders` for a `page`

We have [INC-850](https://dsdmoj.atlassian.net/browse/INC-850) for storing it in the DB which as you commented we may not have sufficient information for. I think we'll know everything we need once we finish [INC-816](https://dsdmoj.atlassian.net/browse/INC-816), so we can park this until then?

ushkarev: yeah i'm totes ok with a lot of the planned functionality to be part-way-there cos so much of it depends on other parts being done

ushkarev (approved):
`overdueCount` uses too narrow a query to count overdue reviews, but that's probs ok for now. there are many other changes still needed to feed into this number to make it right anyway

PR #240, opened 2022-11-03
INC-819: Next review date for new prisoners

xoen:
New prisoners should be reviewed within 3 months (it's within 1 month for "Young People" but that's a separate ticket)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/OffenderSearch.kt
```diff
@@ -1,13 +1,16 @@
 package uk.gov.justice.digital.hmpps.incentivesapi.dto
 
+import java.time.LocalDate
+
 data class OffenderSearchPrisoner(
   val prisonerNumber: String,
   val bookingId: String,
   val firstName: String,
-  val middleNames: String,
+  val middleNames: String?,
```
neilmendum, line 9: This fixes the issue that [Andy flagged](https://mojdt.slack.com/archives/C03T6P453CM/p1667392627468919)?

neilmendum (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/OffenderSearch.kt
xoen: yeah, the middle names can be `null` for some reason. We could potentially remove as it's not used at the moment but I didn't as 1) it was there 2) we have first/last name so I thought I'd keep.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateService.kt
```diff
@@ -49,8 +60,12 @@ class NextReviewDateService(private val input: NextReviewDateInput) {
...
 
   private fun wasConfirmedBasic(): Boolean {
-    val (iepDetails) = input
+    val iepDetails = input.iepDetails
 
     return isOnBasic() && iepDetails.size >= 2 && iepDetails[1].iepLevel == BASIC
   }
+
+  private fun isNewPrisoner(): Boolean {
+    return input.iepDetails.isEmpty()
```
ushkarev, line 69: i have a sneaking suspicion that we might want to instead check that `iepDetails` only has not-real-review-type rows cos there will be an automatic one created at induction

xoen: Yeah, I think this is simple and nice at the moment but I saw another ticket about only considering "real reviews".

Ultimately, we're iterating this to get to a more accurate next review date.

ushkarev: hmm can't see how `iepDetails` is could ever be empty. given that this is calculating the next review date to put into a response that returns someone's *current incentive level*, the code path can't reach here without having found _some_ records from the past.

PR #241, opened 2022-11-03
INC-853: Added "young people" exception to new prisoners rule

xoen:
New prisoners needs to be generally reviewed within 3 months unless they're
"young people" (under the age of 18), in that case they need to be reviewed
within one month.

**IMPORTANT**: The next review date is likely to be calculated at a later
time from their reception date. Therefore the rule uses the age at the
reception date instead of the person age when we run the code.

e.g. if someone was 17yo on their reception date but we calculate the next
review date months later when that person is no longer a "young person"
their next review date should still be reception date + 1 month; Because
they were under the age of 18 on the reception date.

Ticket: [INC-853](https://dsdmoj.atlassian.net/browse/INC-853)


neilmendum (approved):
LGTM

PR #243, opened 2022-11-04
INC-892: Make test cases more consistent

ushkarev:
prison-api and offender search should return the same offenders in test stubs/mocks


ushkarev:
this has been merged into #244

PR #244, opened 2022-11-04
INC-900 Switch to offender search for old review

Mjwillis:
Signed-off-by: mjwillis <michael.willis@digital.justice.gov.uk>


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/OffenderSearch.kt
```diff
@@ -4,9 +4,9 @@ import java.time.LocalDate
 
 data class OffenderSearchPrisoner(
   val prisonerNumber: String,
-  val bookingId: String,
+  val bookingId: Long,
```
ushkarev, line 7: offender search returns this as a string, will jakson (or whatever) correctly convert it

Mjwillis: yes -it does

xoen (approved):
LGTM

PR #248, opened 2022-11-08
INC-903: Tidy up code

ushkarev:
Mostly: `internal` visibility modifier is effectively "public" because this only contains one module, so it might as well be removed or replaced with something more private.
Minor tidy-up: removal of unnecessary suppressions, import fix, more consistent use of object mappers in tests.


src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/integration/SqsIntegrationTestBase.kt
```diff
@@ -49,6 +48,7 @@ class SqsIntegrationTestBase : IntegrationTestBase() {
   companion object {
     private val localStackContainer = LocalStackContainer.instance
 
+    @Suppress("unused")
     @JvmStatic
     @DynamicPropertySource
     fun testcontainers(registry: DynamicPropertyRegistry) {
```
ushkarev, line 54: maybe this entire companion object can be deleted? tests still pass without it and i can’t tell what it’s supposed to do

xoen (approved):

PR #254, opened 2022-11-11
INC-908: Title-case prisoner names when returning reviews information

ushkarev:
…to be consistent with other DPS services (NOMIS source names are always uppercase) even though occasionally it won't produce the right casing (e.g. MacLean would show as Maclean instead)


neilmendum (approved):
LGTM

PR #255, opened 2022-11-11
INC-909: Fall back to standard level first when on transfer the target level is unavailable and there's no default set

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
```diff
@@ -20,8 +20,10 @@ class IepLevelService(
 
   fun chooseDefaultLevel(prisonId: String, prisonLevels: List<IepLevel>): String {
     val activePrisonLevels = prisonLevels.filter(IepLevel::active)
-    return activePrisonLevels.find(IepLevel::default)?.iepLevel
-      // NOMIS traditionally takes the first if no default was found
```
ushkarev, line 24: apparently, NOMIS would select *no* level. but falling back to standard first makes more sense policy-wise. regardless: this is an extreme edge case

neilmendum (approved):
LGTM

PR #261, opened 2022-11-14
INC-921: Fix situation during migration where assigned living unit is not known

ushkarev:
`assignedLivingUnitId` can be 0 which is not a location that can be looked up in prison-api (cf. `GET /api/bookings/{bookingId}`)


andymarke (approved):
LGTM

neilmendum (approved):

PR #263, opened 2022-11-15
INC-817: Store next review date in the database

xoen:
- creates `next_review_date` database table with `booking_id`/`next_review_date` columns
- mechanism to update next review date for one/multiple prisoners
- mechanism to get next review date for one/multiple prisoners
  - this tries to get them from the database first
  - update these in the database if missing
  - return the next review dates
- create a new review triggers the update of the next review date in the database
  - this includes admissions, transfers, merges, etc...

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/NextReviewDate.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/NextReviewDate.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.incentivesapi.jpa
 
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.data.annotation.Id
+import org.springframework.data.annotation.ReadOnlyProperty
 import org.springframework.data.annotation.Transient
 import org.springframework.data.domain.Persistable
 import java.time.LocalDate
@@ -13,6 +14,7 @@ data class NextReviewDate(
 
   val nextReviewDate: LocalDate,
 
+  @ReadOnlyProperty
   val whenCreated: LocalDateTime? = null,
 
   @Transient
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/NextReviewDate.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/NextReviewDate.kt
@@ -1,0 +1,24 @@
+package uk.gov.justice.digital.hmpps.incentivesapi.jpa
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.data.annotation.Id
+import org.springframework.data.annotation.Transient
+import org.springframework.data.domain.Persistable
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+data class NextReviewDate(
+  @Id
+  val bookingId: Long,
+
+  val nextReviewDate: LocalDate,
+
+  val whenCreated: LocalDateTime? = null,
+
+  @Transient
+  @Value("false")
+  val new: Boolean = true,
+) : Persistable<Long> {
+  override fun getId(): Long = bookingId
+  override fun isNew(): Boolean = new
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
@@ -6,12 +6,10 @@ import kotlinx.coroutines.flow.toList
 import org.apache.commons.text.WordUtils
 import org.springframework.stereotype.Service
 import org.springframework.web.reactive.function.client.WebClientResponseException.NotFound
-import uk.gov.justice.digital.hmpps.incentivesapi.config.ListOfDataNotFoundException
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.IncentiveReview
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.IncentiveReviewResponse
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.OffenderSearchPrisoner
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.prisonapi.CaseNoteUsage
-import uk.gov.justice.digital.hmpps.incentivesapi.jpa.repository.PrisonerIepLevelRepository
 import java.time.Clock
 import java.time.LocalDate
 
@@ -19,7 +17,7 @@ import java.time.LocalDate
 class IncentiveReviewsService(
   private val offenderSearchService: OffenderSearchService,
   private val prisonApiService: PrisonApiService,
-  private val prisonerIepLevelRepository: PrisonerIepLevelRepository,
+  private val nextReviewDateGetterService: NextReviewDateGetterService,
   private val clock: Clock,
 ) {
   /**
@@ -47,14 +45,8 @@ class IncentiveReviewsService(
     val deferredPositiveCaseNotesInLast3Months = async { getCaseNoteUsage("POS", "IEP_ENC", prisonerNumbers) }
     val deferredNegativeCaseNotesInLast3Months = async { getCaseNoteUsage("NEG", "IEP_WARN", prisonerNumbers) }
 
-    val bookingIds = offenders.content.map(OffenderSearchPrisoner::bookingId)
-    val nextReviewDates = getNextReviewDatesForOffenders(bookingIds)
-
-    val bookingIdsMissingIepReviews = bookingIds.toSet() subtract nextReviewDates.keys
-    if (bookingIdsMissingIepReviews.isNotEmpty()) {
-      throw ListOfDataNotFoundException(bookingIdsMissingIepReviews)
-    }
-    val overdueCount = nextReviewDates.map { r -> r.value.nextReviewDate }.count { it.isBefore(LocalDate.now(clock)) }
+    val nextReviewDates = nextReviewDateGetterService.getMany(offenders.content)
+    val overdueCount = nextReviewDates.values.count { it.isBefore(LocalDate.now(clock)) }
 
     val positiveCaseNotesInLast3Months = deferredPositiveCaseNotesInLast3Months.await()
     val negativeCaseNotesInLast3Months = deferredNegativeCaseNotesInLast3Months.await()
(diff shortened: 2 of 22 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateUpdaterService.kt
```diff
@@ -0,0 +1,86 @@
...
+        )
+      ).calculate()
+
+      NextReviewDate(
+        bookingId = bookingId,
+        nextReviewDate = nextReviewDate,
+      )
+    }
+
+    nextReviewDateRepository.saveAll(nextReviewDateRecords)
```
ushkarev: tho the docs are extremely sparse, maybe `nextReviewDateRecords = nextReviewDateRepository.saveAll(nextReviewDateRecords).toList()` could pick up things like updated `when_created`. side benefit is that `toList` collects.

xoen: 👌 however in this context I don't really care about the result of `saveAll()` (aside from the fact it actually _saves_ things! 😝) so `collect()` should do :)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
```diff
@@ -47,14 +45,8 @@ class IncentiveReviewsService(
...
 
-    val bookingIds = offenders.content.map(OffenderSearchPrisoner::bookingId)
-    val nextReviewDates = getNextReviewDatesForOffenders(bookingIds)
-
-    val bookingIdsMissingIepReviews = bookingIds.toSet() subtract nextReviewDates.keys
-    if (bookingIdsMissingIepReviews.isNotEmpty()) {
-      throw ListOfDataNotFoundException(bookingIdsMissingIepReviews)
-    }
-    val overdueCount = nextReviewDates.map { r -> r.value.nextReviewDate }.count { it.isBefore(LocalDate.now(clock)) }
+    val nextReviewDates = nextReviewDateGetterService.getMany(offenders.content)
```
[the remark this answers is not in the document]
xoen: Well, if for whatever reason it fails it wouldn't get to to this point but assuming the getter does not fail (e.g. no DB query errors etc) it would return a `Map` with all the dates for all the passed offenders.

That's why:
1. I've removed this exception
2. I've retained `!!` when I'm getting data out of the map, as `nextReviewDates[bookingId]` should always return a `LocalDate` for all the offenders

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateUpdaterService.kt
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    val offendersMap = offenders.associateBy(OffenderSearchPrisoner::bookingId)
+    val bookingIds = offendersMap.keys.toList()
+
+    val iepLevels: Map<String, IepLevel> = prisonApiService.getIncentiveLevels()
+
+    // NOTE: This is to account for bookingIds potentially without any review record
+    val bookingIdsNoReviews = bookingIds.associateWith { emptyList<PrisonerIepLevel>() }
+    val reviewsMap = bookingIdsNoReviews + prisonerIepLevelRepository.findAllByBookingIdInOrderByReviewTimeDesc(bookingIds)
```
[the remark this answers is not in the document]
xoen: It wouldn't work: To calculate the next review date it needs to get all the latest reviews, not just the one with `current = true`. 

`findAllByBookingIdInAndCurrentIsTrueOrderByReviewTimeDesc()` only returns the _current_ reviews for the given `bookingIds`, which is not what I want in this context.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateGetterService.kt
```diff
@@ -0,0 +1,53 @@
...
+   * Tries to get next review dates from the database. For bookingIds without a next review date record it calculates
+   * and persist the next review dates before returning it with the ones already present.
+   *
+   * @param offenders is the list of offenders to get the next review dates for
+   *
+   * @return a Map with bookingIds as keys and next review dates as values
+   */
+  suspend fun getMany(offenders: List<OffenderSearchPrisoner>): Map<Long, LocalDate> {
+    val bookingIds = offenders.map(OffenderSearchPrisoner::bookingId)
+    var existingNextReviewDates = nextReviewDateRepository.findAllById(bookingIds).toList().toMap()
```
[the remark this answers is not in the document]
xoen: Yes that's right! I was originally trying to do something else where I was adding to "this" list but I changed the code but I forgot to change. I'll change to `val`, thanks!

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/NextReviewDate.kt
```diff
@@ -0,0 +1,26 @@
...
+import java.time.LocalDateTime
+
+data class NextReviewDate(
+  @Id
+  val bookingId: Long,
+
+  val nextReviewDate: LocalDate,
+
+  @ReadOnlyProperty
+  val whenCreated: LocalDateTime? = null,
```
neilmendum, line 18: Wonder if including a field for _updated at_ could be useful, then we could see when a record was last updated

neilmendum: It's also possible for Spring Data to manage these fields for us with annotations - not required but you may like to 😄 
 - `@CreatedBy` for `whenCreated`
 - `@LastModifiedDate` for `whenUpdated` (if we decide to add it)

That would require adding
 - `@EntityListeners(AuditingEntityListener::class)` to this class
 - `@EnableJpaAuditing` to one of the `@Configuration` bean classes

You can read more about them [here](https://www.baeldung.com/database-auditing-jpa)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/adapters.kt
```diff
@@ -0,0 +1,32 @@
...
+    reviewType = reviewType,
+    comments = commentText,
+    userId = reviewedBy,
+    locationId = locationId,
+    prisonerNumber = prisonerNumber,
+    auditModuleName = "INCENTIVES_API",
+  )
+
+fun List<PrisonerIepLevel>.toIepDetails(iepLevels: Map<String, IepLevel>): List<IepDetail> {
+  return this.map { review -> review.translate(iepLevels) }
```
neilmendum, line 27: I'm not sure you need `this` here

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/adapters.kt
```diff
@@ -0,0 +1,32 @@
...
+}
+
+fun List<NextReviewDate>.toMap(): Map<Long, LocalDate> {
```
neilmendum, line 30: Could name this `toMapByBookingId` so folks know the shape of the collection returned?

neilmendum (commented):
Looks good! made minor comments and suggestions

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/adapters.kt
```diff
@@ -0,0 +1,32 @@
...
+  return this.associate { record -> Pair(record.bookingId, record.nextReviewDate) }
```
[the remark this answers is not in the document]
xoen: Nice, done ✅

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateGetterService.kt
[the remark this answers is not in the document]
xoen: sorted here: 6eca2ca80c14e3442dd24e60d4a9def990ee9541

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateUpdaterService.kt
xoen: Alright, I'm changing this and the tests to use the returned flow 👍

xoen: Done here: 7622c201d99891938233b5af56ab8507752b2397

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/NextReviewDate.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.data.annotation.Id
+import org.springframework.data.annotation.Transient
+import org.springframework.data.domain.Persistable
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+data class NextReviewDate(
+  @Id
+  val bookingId: Long,
```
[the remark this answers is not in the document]
xoen: Not sure it would help much and it would be another thing to be sure we keep up-to-date/in-sync with Prison API which I'd personally avoid.

It would be more useful to add support for the Offender Search `POST` endpoint that gets offender by bookingIds as I could avoid few requests here and there, but that has slightly different role requirements so I didn't want to add it as part of this.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/adapters.kt
xoen: Could do, I'm also gonna rename `.translate()` to `toIepDetail()` as it still puzzles me when I look at it for a similar reason then.

xoen: Done here: 5bad1d239e9e7b216bcd3eac016a3d33fab75420

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/NextReviewDate.kt
xoen: 🤔 I think `@CreatedBy` is to persist the user rather than the date.

It's good to know we have these available but I think for this table/class is probably overkill, especially as changes mostly happen in the background rather than because of a direct user change, so I'll probably keep it simple and leave it as it is :)

neilmendum: Oops - I meant `@CreatedDate`

ushkarev (approved):

PR #269, opened 2022-11-17
INC-928: Allow flakey test cases more time to await setup conditions

ushkarev:
One of the conditions in lines 63-65 occasionally fails in CI before the default 10s timeout is up. Less often for 93-95. Might as well allow more time for all conditions.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/integration/service/PrisonOffenderEventListenerIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/integration/service/PrisonOffenderEventListenerIntTest.kt
@@ -24,6 +24,8 @@ import java.time.Instant
 import java.time.LocalDateTime
 
 class PrisonOffenderEventListenerIntTest : SqsIntegrationTestBase() {
+  private val awaitAtMost30Secs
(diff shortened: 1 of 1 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/integration/service/PrisonOffenderEventListenerIntTest.kt
```diff
@@ -24,6 +24,8 @@ import java.time.Instant
 import java.time.LocalDateTime
 
 class PrisonOffenderEventListenerIntTest : SqsIntegrationTestBase() {
+  private val awaitAtMost30Secs
+    get() = await.atMost(Duration.ofSeconds(30))
```
ushkarev, line 28: this funny setup is how `await` is defined, so following their example
https://github.com/awaitility/awaitility/blob/a56dce91b2abff03beb4773180ec6e5edb0117d7/awaitility-kotlin/src/main/kotlin/org/awaitility/kotlin/AwaitilityKt.kt#L29-L34

ushkarev:
lol CI didn't fail _this_ time so clearly everything must be fixed 🤣

neilmendum (approved):
LGTM

PR #270, opened 2022-11-17
INC-843: Allow api clients to change sorting order of reviews

ushkarev:


xoen (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveReviewsResource.kt
```diff
@@ -57,6 +59,14 @@ class IncentiveReviewsResource(private val incentiveReviewsService: IncentiveRev
     @PathVariable
     levelCode: String,
 
+    @Schema(description = "Sort reviews by", required = false, defaultValue = "NEXT_REVIEW_DATE", example = "NEXT_REVIEW_DATE", allowableValues = ["NEXT_REVIEW_DATE"])
+    @RequestParam(required = false)
+    sort: IncentiveReviewSort? = null,
```
xoen, line 69: Minor one on the name, shouldn't be reviews (plural) for consistency with other things related to this reviews endpoint? (but maybe there is a reason why this is singular)

ushkarev: Are you referring to the class name `IncentiveReviewSort`? I think it’s more natural English to call something that sorts reviews (plural) a "review sort" (no plural on the noun adjunct); like "card sort". The annoying side effect is, as you point out, that the class name prefix differs between classes… not sure which matters more 🤪

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IncentiveReview.kt
```diff
@@ -24,7 +24,7 @@ data class IncentiveReview(
   val negativeBehaviours: Int,
 
   @Schema(description = "Whether the prisoner has an ACCT open alert", example = "true")
-  val acctOpenStatus: Boolean,
+  val acctStatus: Boolean,
```
xoen, line 27: What was the rational for this rename? "acctOpenStatus = true" kinda reads better (because it's open or closed) than "acctStatus = true".

Maybe being on an ACCT status is how they call having an ACCT alert?

xoen (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IncentiveReview.kt
ushkarev: no good reason 😉. it's what was in the UI app contract. not sure either are the "best". maybe `hasAcctOpen`????

neilmendum (approved):
LGTM

PR #271, opened 2022-11-17
Fix type error in health checks

ushkarev:
…because the HTTP response is nullable so retrieving `.statusCode` is fallible: better to handle the exception explicitly:

> > Task :compileKotlin
> w: hmpps-incentives-api/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/health/HealthCheck.kt: (16, 42): Unsafe use of a nullable receiver of type ResponseEntity<String!>?

i.e. `responseEntity.statusCode` will throw a null pointer exception which is less clear

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/health/HealthCheck.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/health/HealthCheck.kt
@@ -12,7 +12,7 @@ abstract class HealthCheck(private val webClient: WebClient, private val timeout
       .uri("/health/ping")
       .retrieve()
       .toEntity(String::class.java)
-      .block(timeout)
+      .block(timeout) ?: throw Exception("/health/ping did not return a response")
     Health.up().withDetail("HttpStatus", responseEntity.statusCode).build()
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/health/HealthCheck.kt
```diff
@@ -12,7 +12,7 @@ abstract class HealthCheck(private val webClient: WebClient, private val timeout
       .uri("/health/ping")
       .retrieve()
       .toEntity(String::class.java)
-      .block(timeout)
+      .block(timeout) ?: throw Exception("/health/ping did not return a response")
```
neilmendum, line 15: Wonder if we can cover this case in `HealthCheckTest`?

neilmendum: Good point. I'm not sure sure either how to test it 😆 Good to include but don't think there's much benefit of labouring over adding a test so will approve & merge

neilmendum (approved):

PR #272, opened 2022-11-18
INC-901: Update next review date on alerts changes

xoen:
- When a prisoner is updated and if the alerts changed update the next review date (see below)
- Fix incentives' reviews domain events description always being "An IEP review has been added", even tho maybe a review was actually patched or deleted (although admittedly these should happen less frequently)
- Better names for methods

**NOTE**: This is the processing side of the ticket, we need to [update the CP terraform resource](https://github.com/ministryofjustice/cloud-platform-environments/blob/main/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-domain-events-dev/resources/hmpps-incentives.tf#L118) to also get `prisoner-offender-search.prisoner.updated` domain events.

###  Notes on domain event used

Currently based on `prisoner-offender-search.prisoner.updated` event (`Domain` topic).
This event is published every time a prisoner is updated, this means we need to check if the `categoriesChanged` field lists `ALERTS` as changed. Furthermore we should really re-calculate the next review date when ACCT alerts change (`alertCode == "HA"`).

We're planning to add a `prisoner-offender-search.prisoner.alerts-updated` domain event which would be published only when the alerts changes. This would also include the `alertCode` in the `additionalInformation`, which means we should receive less messages and update the next review date only when ACCT status changed.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
@@ -269,6 +269,26 @@ class PrisonerIepLevelReviewService(
       }
     }
 
+  @Transactional
+  suspend fun processPrisonerUpdatedEvent(prisonOffenderEvent: HMPPSDomainEvent) {
+    val alertsChanged: Boolean = prisonOffenderEvent.additionalInformation.categoriesChanged?.contains("ALERTS") ?: false
+
+    if (alertsChanged) {
+      updateNextReviewDate(prisonOffenderEvent)
+    } else {
+      log.debug("Ignoring 'prisoner-offender-search.prisoner.updated' event, alerts didn't change: categoriesChanged = ${prisonOffenderEvent.additionalInformation.categoriesChanged}")
+    }
+  }
+
+  private suspend fun updateNextReviewDate(prisonOffenderEvent: HMPPSDomainEvent) {
+    prisonOffenderEvent.additionalInformation.nomsNumber?.let { prisonerNumber ->
+      val offender = offenderSearchService.getOffender(prisonerNumber)
+      nextReviewDateUpdaterService.updateMany(listOf(offender))
+    } ?: run {
+      log.warn("prisonerNumber null for prisonOffenderEvent: $prisonOffenderEvent ")
+    }
+  }
+
   private suspend fun createIepForReceivedPrisoner(prisonOffenderEvent: HMPPSDomainEvent, reviewType: ReviewType) {
     prisonOffenderEvent.additionalInformation.nomsNumber?.let {
       val prisonerInfo = prisonApiService.getPrisonerInfo(it, true)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
@@ -276,7 +276,7 @@ class PrisonerIepLevelReviewService(
     if (alertsChanged) {
       updateNextReviewDate(prisonOffenderEvent)
     } else {
-      log.debug("Ignoring 'prisoner-offender-search.prisoner.updated' event, alerts didn't change: categoriesChanged = ${prisonOffenderEvent.additionalInformation.categoriesChanged}")
+      log.debug("Ignoring 'prisoner-offender-search.prisoner.updated' event, alerts didn't change: prisonerNumber = ${prisonOffenderEvent.additionalInformation.nomsNumber}, categoriesChanged = ${prisonOffenderEvent.additionalInformation.categoriesChanged}")
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -269,6 +269,34 @@ class PrisonerIepLevelReviewService(
...
     }
 
+  @Transactional
+  suspend fun processPrisonerUpdatedEvent(prisonOffenderEvent: HMPPSDomainEvent) {
+    val alertsChanged: Boolean = prisonOffenderEvent.additionalInformation.categoriesChanged?.contains("ALERTS") ?: false
+
+    if (alertsChanged) {
+      updateNextReviewDate(prisonOffenderEvent)
+    } else {
+      log.debug("Ignoring 'prisoner-offender-search.prisoner.updated' event, alerts didn't change: categoriesChanged = ${prisonOffenderEvent.additionalInformation.categoriesChanged}")
```
neilmendum, line 279: Could be worth also including `additionalInformation.nomsNumber` in the log statement?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -269,6 +269,34 @@ class PrisonerIepLevelReviewService(
...
+
+      snsService.publishDomainEvent(
+        id = offender.bookingId,
+        nomsNumber = offender.prisonerNumber,
+        occurredAt = LocalDateTime.now(clock),
+        eventType = IncentivesDomainEventType.PRISONER_NEXT_REVIEW_DATE_UPDATED,
+        description = "A prisoner next review date was updated",
+      )
+    } ?: run {
+      log.warn("prisonerNumber null for prisonOffenderEvent: $prisonOffenderEvent ")
```
neilmendum, line 296: You could additionally mention the result of it being null? For example `prisonerNumber null for prisonOffenderEvent: $prisonOffenderEvent so could not update next review date`

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonOffenderEventListenerTest.kt
```diff
@@ -68,6 +68,17 @@ class PrisonOffenderEventListenerTest {
     }
   }
 
+  @Test
+  fun `process prisoner updated message`(): Unit = runBlocking {
```
neilmendum, line 72: Could also add a test to `PrisonOffenderEventListenerIntTest` for this? That would validate the SQS listener (running in Localstack) and database is updated

neilmendum (commented):
Looks good - added some comments

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
xoen: Good idea, I'll add the reason to the warning 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
xoen: I'll add 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -426,12 +446,21 @@ class PrisonerIepLevelReviewService(
...
     eventType: IncentivesDomainEventType,
   ) {
     iepDetail.id?.let {
-      snsService.sendIepReviewEvent(iepDetail.id, iepDetail.prisonerNumber ?: "N/A", iepDetail.iepTime, eventType)
+      val description: String = when (eventType) {
+        IncentivesDomainEventType.IEP_REVIEW_INSERTED -> "An IEP review has been added"
+        IncentivesDomainEventType.IEP_REVIEW_UPDATED -> "An IEP review has been updated"
+        IncentivesDomainEventType.IEP_REVIEW_DELETED -> "An IEP review has been deleted"
+        else -> {
+          throw IllegalArgumentException("Tried to publish a review event with a non-review event type: $eventType")
```
ushkarev, line 459: as there are no other variants of `IncentivesDomainEventType`, is this `else` block there to indicate an error in implementation?

xoen: Pretty much, it's basically to highlight our code tried to publish a review domain event with an unexpected domain event type.

ushkarev (approved):
ok from what i can tell, but not totally confident about approving 😉

neilmendum (approved):
LGTM

PR #288, opened 2022-11-24
INC-940: Investigate flakey tests

ushkarev:
There's also a [CircleCI orb for using localstack](https://docs.localstack.cloud/ci/circle-ci/) but that might be harder to use.


neilmendum (approved):
LGTM

PR #289, opened 2022-11-25
INC-946: Maintenance

ushkarev:
• Minor read-me change
• [Upgrade gradle](https://docs.gradle.org/7.6/release-notes.html) which supports java v19 for running itself (the project toolchain was already java v19); the kotlin plugin however still only supports java v18 bytecode
• Stable fake SQS topic ARN for domain events in tests and when running locally


ushkarev:
During tests and when running locally, the SNS topic ARN is randomly generated: is this strictly necessary? I wondered if it might be easier to hard-code an obviously-fake ARN…

https://github.com/ministryofjustice/hmpps-incentives-api/blob/bff9b747f91048a9594b8590c35d59b639ceef4a/src/test/resources/application-test.yml#L50
https://github.com/ministryofjustice/hmpps-incentives-api/blob/bff9b747f91048a9594b8590c35d59b639ceef4a/src/main/resources/application-localstack.yml#L14

xoen (approved):

PR #291, opened 2022-11-25
INC-942 Merge process incorrectly moves booking data when it should COPY

Mjwillis:
As detailed in https://dsdmoj.atlassian.net/browse/INC-939 

The change needed is to copy incentive data from the old bookings that have been merged into the current booking.  The existing incentive details for a previous booking are left alone.  

This replaces what NOMIS is doing upon a merge


src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewServiceTest.kt
```diff
@@ -666,48 +667,62 @@ class PrisonerIepLevelReviewServiceTest {
...
+            oldReview2,
+            oldReview1
           )
         )
       prisonerIepLevelReviewService.mergedPrisonerDetails(prisonMergeEvent)
 
+      verify(prisonerIepLevelRepository).save(eq(newReview.copy(prisonerNumber = "A1244AB")))
+      verify(prisonerIepLevelRepository).save(eq(oldReview2.copy(bookingId = 1234567L, id = 0L, current = false)))
+      verify(prisonerIepLevelRepository).save(eq(oldReview1.copy(bookingId = 1234567L, id = 0L, current = false)))
+
```
xoen, line 725: So:
1. the new reviews with the new/incorrect prisoner number are updated to have the existing/correct prisoner number
2. the old reviews with the old booking id and existing/correct prisoner number are copied (but not updated because of the `id = 0L`, right?) into reviews with existing/correct prisoner number but also new booking Id so that history for merged prisoner includes old reviews

Is that correct? Makes more sense now :)

Mjwillis: yes

Mjwillis:
@xoen I've resolved the test issue - it was a real bug it was highlighting

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewServiceTest.kt
```diff
@@ -657,48 +658,62 @@ class PrisonerIepLevelReviewServiceTest {
...
-              bookingId = 555555L,
-              prisonId = "LEI",
-              locationId = "LEI-1-1-001",
-              reviewedBy = "TEST_STAFF1",
-              iepCode = "STD",
-              current = true,
-              reviewTime = LocalDateTime.now().minusDays(100)
-            )
+            oldReview2,
+            oldReview1
```
xoen, line 708: I don't think it matters for this test but the order would be `flowOf(oldReview1, oldReview2)` because the query sort by review time desc and `oldReview1` is more recent (100 days old) than `oldReview2` (200 days old)

...but as I said, it shouldn't impact the test as the merge copy all of them reviews with the old `prisonerNumber` regardless :)

xoen (dismissed):

xoen (approved):

PR #292, opened 2022-11-28
INC-949: Migration endpoint don't update next review date

xoen:
During the migration a lot of intermediate reviews are created. It doesn't make much sense to update the next review date every time a potentially old review is created.

This has also 2 performance benefits:
- we avoid making more requests when handling migration requests which makes the migration faster
- when we'll publish domain events on next review date change these will trigger a reindex of the prisoner in ES/Offender Search API: we don't want to trigger a reindex for each of the intermediate/old reviews being migrated from NOMIS

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
@@ -153,14 +153,14 @@ class PrisonerIepLevelReviewService(
       )
     )
 
-    nextReviewDateUpdaterService.update(review.bookingId)
-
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -153,15 +153,15 @@ class PrisonerIepLevelReviewService(
       )
     )
 
-    nextReviewDateUpdaterService.update(review.bookingId)
```
xoen, line 156: This was done inside the `persistSyncPostRequest()` method. This method is used during the handling of both sync POST requests and also migration requests.

I've moved this call out so that it now only happen during the handling of sync POST requests (`handleSyncPostIepReviewRequest()` method).

ushkarev (approved):

PR #295, opened 2022-11-29
INC-927: Publish domain event when next review date changes

xoen:
Publishes a `incentives.prisoner.next-review-date-changed` domain event when the next review date record for a prisoner is updated.

The domain event also includes the new next review date.

NOTE: The event is only published when we can be sure this has changed, e.g. there was a previous record and the value of the next review date has changed. This means when a next review date was not present in the DB we don't publish a domain event. This was suggested by Andy and it should help avoiding unnecessary OffenderSearch API re-indexes.
It also means if we were to truncate the `next_review_date` DB table we'd avoid potentially duplicated messages.


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateUpdaterService.kt
```diff
@@ -84,6 +87,25 @@ class NextReviewDateUpdaterService(
...
+    }
+
+    // TODO: It should be possible to publish all these domain events concurrently rather than sequentially
+    bookingIdsChanged.forEach { bookingId ->
+      snsService.publishDomainEvent(
+        id = bookingId,
+        nomsNumber = offendersMap[bookingId]!!.prisonerNumber,
+        occurredAt = LocalDateTime.now(clock),
+        eventType = IncentivesDomainEventType.PRISONER_NEXT_REVIEW_DATE_CHANGED,
+        description = "A prisoner next review date has changed",
```
ushkarev: maybe "A prisoner's next _incentive_ review date has changed"

xoen: I'll change as it's good to give more context 👍 (it's kinda implicit because of the full domain event type but it's good to be more explicit nevertheless)

ushkarev: the whole description is implicit in the event name, but i figure the description might as well be a sentence to explain it

xoen: Done here: https://github.com/ministryofjustice/hmpps-incentives-api/pull/295/commits/7ce33374de254b32ad57008e0fd114a060b4949d

ushkarev (approved):

PR #297, opened 2022-11-30
INC-957: Don't make requests to Offender Search API to calculate next review date

xoen:
Problem
--------

When Offender Search tries to re-index a prisoner which has just been created/admitted to prison the Offender Search API only contains the basic prisoner info before their admission (so no `bookingId`, no reception date, no alerts, no status etc).

Offender Search API tries to get incentive level/next review date for the prisoner from the Incentives API.
Before this change the calculation to get the next review date was triggering a request to a yet-to-be-updated Offender Search API causing a cycle/issue.

Solution
--------

When calculating/updating the next review date for a prisoner get the information required from Prison API's `GET /api/bookings/:bookingId?extraInfo=true`. The response format is very _similar_ (but not exactly the same!) to the response from Offender Search API and crucially:
1. contains all the information needed by the next review date calculation (date of birth, reception date, alerts) 
2. shouldn't be much slower than the Offender Search API request (and we would make only one HTTP request instead of two)

Note that the Offender Search reindexing would make a request to Incentives' `GET /reviews/booking/:bookingId` so by the point we process this request the prisoner was admitted (as they have a `bookingId` which they get on admission!) and they'd have a `receptionDate` and `alerts` (although probably an empty list/no alerts)

See Slack thread for some more context: https://mojdt.slack.com/archives/C03T6P453CM/p1669654105323679

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateUpdaterService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateUpdaterService.kt
@@ -95,8 +95,6 @@ class NextReviewDateUpdaterService(
 
     publishDomainEvents(bookingIdsChanged, offendersMap, nextReviewDatesAfterUpdate)
 
-    println("nextReviewDatesAfterUpdate = $nextReviewDatesAfterUpdate")
-
     return nextReviewDatesAfterUpdate
   }
 


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/OffenderSearch.kt
```diff
@@ -1,33 +1,22 @@
...
-  val bookingId: Long,
+  override val bookingId: Long,
+  override val prisonerNumber: String,
+  override val dateOfBirth: LocalDate,
+  override val receptionDate: LocalDate,
   val firstName: String,
   val middleNames: String? = null,
   val lastName: String,
-  val status: String,
-  val inOutStatus: String,
```
xoen, line 12: I got rid of some fields we don't actually use:
- `status`
- `inOutStatus`
- `prisonName`
- `cellLocation`
- `locationDescription`
- (we probably don't use `prisonId` but Igor suggested could be useful for debugging)

No point keeping these around unused fields and more often than not unused fields can cause deserialisation issues (when types are not 100% correct)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/prisonapidto.kt
```diff
@@ -15,6 +17,26 @@ data class PrisonerAtLocation(
...
+
+data class PrisonerExtraInfo(
+  override val bookingId: Long,
+  override val dateOfBirth: LocalDate,
+  override val receptionDate: LocalDate,
+  val offenderNo: String,
+  val alerts: List<PrisonerAlert> = emptyList(),
+) : PrisonerInfoForNextReviewDate {
+  override val hasAcctOpen = alerts.any(PrisonerAlert::isOpenAcct)
+  override val prisonerNumber = offenderNo
```
xoen, line 37: Prison API's response has a `offenderNo` field instead of calling it `prisonerNumber` like the Offender Search API, so I'm mapping it `offenderNo ` => `offenderNo` to adhere with the `PrisonerInfoForNextReviewDate` interface.

andymarke (approved):
lgtm

Mjwillis (approved):
LGTM

PR #298, opened 2022-12-01
INC-961: Enable sync to NOMIS in `prod`

ushkarev:


xoen (approved):

xoen:
FYI: Marked this PR as draft so that it doesn't accidentally get merged (we're working on a fix for a sync service race condition and more testing is required)

xoen (approved):

PR #299, opened 2022-12-01
INC-962: Fix for sync race condition/duplicated NOMIS reviews

xoen:
Because of the timing between when the sync service tells us about new reviews (via `POST /iep/sync/booking/:bookingId` endpoint) and when we respond/this is stored in the sync mapping service, another domain event is published causing duplicated reviews in NOMIS.

The sync service should really ignore domain events published by this sync POST endpoint because these reviews are already in NOMIS.

We're now adding a `"reason": "USER_CREATED_NOMIS"` (in the usual `additionalInformation`) so that sync service can ignore these domain events and avoid duplicates.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/SnsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/SnsService.kt
@@ -84,5 +84,9 @@ enum class IncentivesDomainEventType(val value: String) {
   PRISONER_NEXT_REVIEW_DATE_CHANGED("incentives.prisoner.next-review-date-changed"),
 }
 
+enum class IepReviewReason {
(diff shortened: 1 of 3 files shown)


ushkarev (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/SnsService.kt
```diff
@@ -84,5 +84,9 @@ enum class IncentivesDomainEventType(val value: String) {
   PRISONER_NEXT_REVIEW_DATE_CHANGED("incentives.prisoner.next-review-date-changed"),
 }
 
+enum class IepReviewReason {
+  USER_CREATED_NOMIS, // NOTE: Used by Syscon sync service to discriminate reviews already synced
```
ushkarev, line 88: 👍  we can fill out the other 2 (+?) later

andymarke (approved):
lgtm

PR #303, opened 2022-12-05
INC-821: Next review date calculation takes readmissions in consideration

xoen:
When someone is re-admitted into prison with the same booking (e.g. because of a recall, remand or potentially other similar cases) a "review" is created with time "READMISSION".

The next review date service checks if the last "review" is a readmission and in that case uses the same rules used for new prisoners to calculate the next review date (so within 3 months from readmission date for adults, within 1 month for young people).

**NOTE**: The Incentives policy only mention recalls but we don't have any reliable way to discriminate between recalls and other kind of readmissions. We think it's OK to treat all readmissions as new prisoners (for the purpose of calculating the next review date) but I'll also check with Steve.


neilmendum (approved):
LGTM

PR #304, opened 2022-12-05
INC-966: Increase `prisoner_iep_level.location_id` column length

xoen:
Some of the SQL INSERT rarely fails because of the fact a location description is longer (up to 240 characters) than the `location_id` column where they're stored (which is currently 30 characters max).

Given we're storing the location description, and as also suggested by Syscon, I'm increasing the length of this column to 240 characters so that it's in line with the maximum possible value.


neilmendum (approved):
LGTM

PR #305, opened 2022-12-06
Use Incentives database as source of truth

xoen:
When getting the history of reviews for a prisoner use the Incentives DB instead of getting it from NOMIS (via Prison API)

NOTE: We can now do this because we migrated all the data and 2-way syncing is enabled.


neilmendum (approved):
LGTM

PR #307, opened 2022-12-07
INC-958: Use new `alerts-updated` domain events to recalculate the next review date

xoen:
Before this we were listening to the `prisoner.updated` domain event and updating the next review date any time any alerts changed. This is not ideal as we receive a lot of events we don't care about and also we potentially re-calculate the next review date when non-ACCT alerts changed.

Since then Andy has added a more specific domain event published when alerts changes (`prisoner-offender-search.prisoner.alerts-updated`), this domain event should be less frequent and it also includes the alert codes of the alerts added/removed meaning we can update the next review date only when an ACCT alert was added/removed.

**NOTE**: The old domain events are no longer processed so this can only be merged once the wiring to subscribe to the new events is working. The CP terraform changes were merged but I haven't seen these new messages in `dev` yet. Once these new messages are processed we can unsubscribe from old messages


neilmendum (approved):
LGTM

PR #314, opened 2022-12-09
INC-970: Fix for 'OUT has no available incentive levels'

xoen:
We're seeing a lot of these exceptions in `dev` (probably caused by the same domain event) but it's very noisy and makes it hard to see other more important problems.

We haven't seen this in production yet but it's feasable for this to happen there as well if, for whatever reason, at the time we process a domain even for a prisoner this is no longer in prison (their current agencyId would be `OUT` which stands for "Outside").

Try to determine the default Incentive level for OUT (which is not a valid prison) would cause an exception so it seems reason to just ignore these domain events (and log a warning message).


xoen:
I have cleared the DQL in `dev` which should reduce the noise for now. If this happens again and/or in production we can investigate further why this is happening and potentially introduce this fix.

PR #322, opened 2022-12-15
Add a simple script to process overdue review stats by prison

ushkarev:
Uses [pandas](https://pandas.pydata.org/) and requires CSV exports from incentives and prisoner money databases.


scripts/next-review-dates-by-prison.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python
```
neilmendum, line 1: Couple of points

- Is there a Jira ticket or Slack thread we can link to which would provide future folks context about why this was added?
- Would be handy to include steps about how to run it
- Just thinking about languages - this is written in Python, while I remembered we had a handful of scripts in [hmpps-incentives-ui/bin](https://github.com/ministryofjustice/hmpps-incentives-ui/tree/main/bin) that are in TypeScript. Would it be possible to write this in TypeScript so we have consistency in languages for Incentives scripts?

neilmendum: Ok fair enough. Thanks for adding the readme. Happy to merge as always handy to have scripts version controlled 👍

neilmendum (approved):
LGTM

PR #328, opened 2023-01-05
INC-1010 Issue with Invalid type in Prisoners in Incentives Review screen (old version)

Mjwillis:
Where a prisoner has just entered prison (since migrating to new service 5 December) - if no REAL review has taken place the prisoner is show in a “Invalid” tab on the review screen.

Change this to put them in a “**No review**” tab, so its clear that no “real review” has yet taken place.

### DIFF
--- a/docker-compose-local.yml
+++ b/docker-compose-local.yml
@@ -6,7 +6,7 @@ services:
       - hmpps
     container_name: localstack
     ports:
-      - "4566-4597:4566-4597"
+      - "4566:4566"
       - "8999:8080"
     environment:
       - DEBUG=${DEBUG- }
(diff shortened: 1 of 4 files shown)


docker-compose-local.yml
```diff
@@ -6,7 +6,7 @@ services:
       - hmpps
     container_name: localstack
     ports:
-      - "4566-4597:4566-4597"
```
Mjwillis, line 9: This is just to stop docker desktop looking weird

docker-compose.yml
```diff
@@ -80,7 +80,7 @@ services:
       - hmpps
     container_name: localstack-incentives-api
     ports:
-      - "4566-4597:4566-4597"
```
Mjwillis, line 83: This is just to stop docker desktop looking weird

andymarke (approved):
lgtm

PR #332, opened 2023-01-06
INC-1026: Load all pages of offender search results

ushkarev:


src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/OffenderSearchServiceTest.kt
```diff
@@ -62,4 +71,93 @@ class OffenderSearchServiceTest {
...
+                )
+              )
+            )
+            .build()
+        } else {
+          ClientResponse.create(HttpStatus.NOT_FOUND).build()
+        }
+        Mono.just(response)
+      }
+      .build()
```
ushkarev, line 97: looks overly complicated but can’t find a simpler way to do it… ideas?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
```diff
@@ -35,7 +35,7 @@ class IncentiveSummaryService(
     sortBy: SortColumn = SortColumn.NAME,
     sortDirection: Sort.Direction = Sort.Direction.ASC
   ): BehaviourSummary = coroutineScope {
-    val prisoners = offenderSearchService.findOffenders(prisonId, locationId, 0, 1000).content.toList()
```
ushkarev, line 38: 1000 was too big a page size: offender search limits it to 200 max

xoen (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/OffenderSearchServiceTest.kt
neilmendum: As discussed - while constructing `WebClient.builder()` is a bit complicated it's potentially neater than attempting to use Mockito to do it, as we're trying to validate the following code in `OffenderSearchService`, I think your approach is ok
```
  page += 1
} while (!pageOfData.last)
```

neilmendum (approved):
LGTM

PR #335, opened 2023-01-09
Update dependencies

ushkarev:


build.gradle.kts
```diff
@@ -81,7 +81,7 @@ java {
 tasks {
   withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
     kotlinOptions {
-      jvmTarget = "18"
+      jvmTarget = "19"
```
xoen, line 84: What changed? I see the Kotlin bootstrap uses 18, version 19 is what we use in the `Dockefile`/when running isn't it?

ushkarev: kotlin 1.8 now lets you compile to 19 class version so i figured it's a sensible idea to have everything as one version. tests pass with and without this setting so i dunno what real effect this has. it's feasible that the kotlin compiler can now make more optimised code for the JRE version it runs under…?

xoen (approved):

PR #339, opened 2023-01-13
INC-1025 Move event publication outside the commit transaction boundary.

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveStoreService.kt
```diff
@@ -0,0 +1,89 @@
+package uk.gov.justice.digital.hmpps.incentivesapi.service
```
Mjwillis, line 1: This is a new service to wrap the insert/updates to incentive data with a Transactional boundary

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -175,16 +155,7 @@ class PrisonerIepLevelReviewService(
...
-    prisonerIepLevelRepository.delete(prisonerIepLevel)
-    nextReviewDateUpdaterService.update(bookingId)
-
-    // If the deleted record had `current=true`, the latest IEP review becomes current
-    if (prisonerIepLevel.current) {
-      // The deleted record was current, set new current to the latest IEP review
-      prisonerIepLevelRepository.findFirstByBookingIdOrderByReviewTimeDesc(bookingId)?.run {
-        prisonerIepLevelRepository.save(this.copy(current = true))
-      }
-    }
```
xoen, line 187: Isn't there a way to wrap this in a transaction without moving it out? e.g. 

```Kotlin
@Transactional
{
  // something
  // something
}
// outside transaction
```

I didn't find anything useful.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -175,16 +155,7 @@ class PrisonerIepLevelReviewService(
...
+    incentiveStoreService.syncIncentiveRecords(prisonerIepLevel, bookingId)
```
xoen, line 158: I don't mind extracting this logic into another service however I suggest the names are consistent to the existing method names to avoid confusion.

E.g. this method handle sync DELETE requests but the new method is called `syncIncentiveRecords ()`. By looking at this line alone I cannot really tell what this method does and I may well think it's updating multiple incentive records?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveStoreService.kt
```diff
@@ -0,0 +1,86 @@
...
+
+  suspend fun updateMergedReviews(
+    reviewsToUpdate: List<PrisonerIepLevel>,
+    remainingBookingId: Long
+  ) {
+    prisonerIepLevelRepository.saveAll(reviewsToUpdate)
+    nextReviewDateUpdaterService.update(remainingBookingId)
+  }
+
+  suspend fun syncIncentiveRecords(
```
xoen, line 50: I would rename to `deleteIncentiveReview()` for consistency/clarity.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveStoreService.kt
```diff
@@ -0,0 +1,86 @@
...
+    // If the deleted record had `current=true`, the latest IEP review becomes current
+    if (prisonerIepLevel.current) {
+      // The deleted record was current, set new current to the latest IEP review
+      prisonerIepLevelRepository.findFirstByBookingIdOrderByReviewTimeDesc(bookingId)?.run {
+        prisonerIepLevelRepository.save(this.copy(current = true))
+      }
+    }
+  }
+
+  suspend fun syncIncentiveReview(
```
xoen, line 66: Would rename to `patchIncentiveReview()` for clarity.

xoen (changes requested):
I get the idea of moving the logic for persisting in a separate service (although it almost feels like this could be in the repository?)

My main issue with this PR is the names of the methods, otherwise it looks good.

I'd rename these (especially the deletion and patch ones) and this would be good to go :)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
Mjwillis: No has to be in a new interface for proxying

Mjwillis:
@xoen Made changes

xoen (approved):
Thanks Michael

PR #342, opened 2023-01-16
INC-988: Cache global list of all incentive levels

xoen:
### Problem
We discovered that quite often getting the global list of all the incentive levels can take several seconds (often taking 30-60 seconds, [see AppInsights query](https://portal.azure.com#@747381f4-e81f-4a43-bf68-ced6a1e14edf/blade/Microsoft_OperationsManagementSuite_Workspace/Logs.ReactView/resourceId/%2Fsubscriptions%2Fa5ddf257-3b21-4ba9-a28c-ab30f751b383%2FresourceGroups%2Fnomisapi-prod-rg%2Fproviders%2FMicrosoft.Insights%2Fcomponents%2Fnomisapi-prod/source/LogsBlade.AnalyticsShareLinkToQuery/q/H4sIAAAAAAAAA0VOuwqDMBTd%252FYqLu8Slo91CKUgpUrpKmhwxoDchV1sK%252Ffimg3Y6cN4OEezA1kOKD71GJJBbk1l8YDrSoa7rnbdTWF3fhQkXM4OahspxjlEqzxa8%252BCekMtGXe4A320nfSGVJJQxZyPbKhdl4FrXhWV%252F7Vt91q2xwkLJQirYeWa2FyK9qMJMgD0hICz3e%252F685Y79cWyJgzgAAAA%253D%253D/timespan/PT4H))

It's not clear why this endpoint can be so slow (it should be fairly simple and quick) but the list of all incentive levels is used in quite a lot of places for a variety of reasons.

### Caching
This list doesn't change very often so it makes sense to cache it for some time.

I've added a simple caching class to keep the value in memory for a day. It should be simple enough for this use case and much faster than making an HTTP request all the times.

**NOTE**: I tried to use Spring Boot's caching mechanism but for some reason I
couldn't get it to work - maybe is the combination of libraries we use.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
@@ -37,7 +37,7 @@ class IepLevelService(
     val levelCodesAvailableInPrison = prisonLevels.filter(IepLevel::active).map(IepLevel::iepLevel).toSet()
 
     data class KnownLevel(val code: String, val available: Boolean)
-    val allKnownLevels = prisonApiService.getIepLevels().toList()
+    val allKnownLevels = prisonApiService.getIepLevels()
       .sortedBy { it.sequence }
       .map {
         KnownLevel(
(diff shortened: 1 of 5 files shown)


neilmendum (approved):
LGTM

PR #353, opened 2023-02-09
INC-1069: Improve next review date logic for readmissions

xoen:
Previous logic was very simple and only checked the last incentive record of a prisoner to determine if it was a readmission.

This doesn't work if the prisoner is transferred before they got reviewed.

The new logic is similar but it allows for newer non-review entries present after the readmission one.

Example:
- `TRANSFER`
- `TRANSFER`
- `READMISSION`
- [...]

The two entries of type `TRANSFER` here are "ignored" (as they're not real reviews). Readmissions are also not real reviews but they're considered when determining whether the prisoner has been readmitted but hasn't been reviewed yet.

The date of the readmission entry is used as a starting date for the calculation of the next review date as expected (+3 months in general but +1 month for young people).


neilmendum (approved):
LGTM

PR #360, opened 2023-02-16
INC-805 - Get behaviour counts based on last review date

Mjwillis:
## What does this do?

- Calls a new endpoint in prison api which returns a list of case note counts group by type/subtype for a list of prisoners but also filters by a `fromDate` which is the last review date
- This is used to then calculate the number of positive (POS) and negative (NEG) case note behaviours raised since the last review.

### Last Review Date Calculation
The last review date is as follows:
- find the latest `Real Review` (REVIEW/MIGRATED) incentive review.  If found - this is the review date
- if no `Real Review` found then use the review date 3 month before today
- if this review date is over 3 month old then make the review date 3 month before today

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/BehaviourService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/BehaviourService.kt
@@ -8,6 +8,8 @@ import java.time.Clock
 import java.time.LocalDateTime
 import java.time.Period
 
+const val DEFAULT_MONTHS = 3L
+
 @Service
 class BehaviourService(
   private val prisonApiService: PrisonApiService,
@@ -42,16 +44,20 @@ class BehaviourService(
     reviews
       .groupBy { it.bookingId }
       .map { review ->
-        val latestReview = review.value.firstOrNull(PrisonerIepLevel::isRealReview) ?: review.value.first()
-        review.key to latestReview
+        val latestReview = review.value.firstOrNull(PrisonerIepLevel::isRealReview)
+        review.key to truncateReviewDate(latestReview?.reviewTime)
       }.associate {
-        it.first to truncateReviewDate(it.second.reviewTime)
+        it.first to it.second
       }
 
-  private fun truncateReviewDate(lastReviewTime: LocalDateTime): LocalDateTime {
+  private fun defaultReviewPeriod() = LocalDateTime.now(clock).minusMonths(DEFAULT_MONTHS)
+
+  private fun truncateReviewDate(lastReviewTime: LocalDateTime?): LocalDateTime {
+    if (lastReviewTime == null) return defaultReviewPeriod()
+
     val today = LocalDateTime.now(clock)
-    return if (Period.between(lastReviewTime.toLocalDate(), today.toLocalDate()).months < 3) { lastReviewTime } else {
-      today.minusMonths(3)
+    return if (Period.between(lastReviewTime.toLocalDate(), today.toLocalDate()).months < DEFAULT_MONTHS) { lastReviewTime } else {
+      defaultReviewPeriod()
     }
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/BehaviourService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/BehaviourService.kt
@@ -1,0 +1,69 @@
+package uk.gov.justice.digital.hmpps.incentivesapi.service
+
+import kotlinx.coroutines.flow.toList
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.incentivesapi.dto.prisonapi.PrisonerCaseNoteByTypeSubType
+import uk.gov.justice.digital.hmpps.incentivesapi.jpa.PrisonerIepLevel
+import java.time.Clock
+import java.time.LocalDateTime
+import java.time.Period
+
+@Service
+class BehaviourService(
+  private val prisonApiService: PrisonApiService,
+  private val clock: Clock,
+) {
+  private fun behaviourCaseNoteMap() = mapOf("POS" to "IEP_ENC", "NEG" to "IEP_WARN")
+
+  suspend fun getBehaviours(reviews: List<PrisonerIepLevel>) =
+    getCaseNoteUsageByLastReviewDate(
+      prisonApiService.retrieveCaseNoteCountsByFromDate(
+        behaviourCaseNoteMap().keys.toList(),
+        getLastRealReviewForOffenders(reviews)
+      ).toList(),
+      behaviourCaseNoteMap(),
+    )
+  private fun getCaseNoteUsageByLastReviewDate(caseNotesByType: List<PrisonerCaseNoteByTypeSubType>, subTypeMap: Map<String, String>) =
+    caseNotesByType
+      .groupBy(PrisonerCaseNoteByTypeSubType::toKey)
+      .map { cn ->
+        CaseNoteSummary(
+          key = cn.key,
+          totalCaseNotes = calcTypeCount(cn.value),
+          numSubTypeCount = calcTypeCount(cn.value.filter { cnc -> cnc.caseNoteSubType == subTypeMap[cn.key.caseNoteType] })
+
+        )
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
```diff
@@ -104,19 +103,19 @@ class IncentiveReviewsService(
     // Count overdue reviews and total reviews by Incentive levels
     val prisonersCounts: MutableMap<String, Int> = mutableMapOf()
     val overdueCounts: MutableMap<String, Int> = mutableMapOf()
-    allReviews.forEach() { review ->
-      val levelCode = review.levelCode
+    allReviews.forEach { review ->
+      val currentReviewLevel = review.levelCode
```
Mjwillis, line 107: The levelCode was shadowing the review level code

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
```diff
@@ -139,24 +138,32 @@ class IncentiveReviewsService(
...
+  private suspend fun getLastRealReviewForOffenders(bookingIds: List<Long>): Map<Long, LocalDateTime> =
+    prisonerIepLevelRepository.findAllByBookingIdInOrderByReviewTimeDesc(bookingIds)
+      .toList()
+      .groupBy { it.bookingId }
+      .map { review ->
+        val latestReview = review.value.firstOrNull(PrisonerIepLevel::isRealReview) ?: review.value.first()
+        review.key to latestReview
+      }.associate {
+        it.first to it.second.reviewTime
+      }
```
Mjwillis, line 150: Will also add in the date check if older than 3 month and default to that

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
```diff
@@ -139,24 +138,32 @@ class IncentiveReviewsService(
...
       .toList()
-      .groupBy(CaseNoteUsage::offenderNo)
+      .groupBy(PrisonerCaseNoteByTypeSubType::toKey)
       .map { cn ->
-        CaseNoteSummary(
-          offenderNo = cn.key,
+        PrisonerCaseNoteSummary(
+          key = cn.key,
           totalCaseNotes = calcTypeCount(cn.value),
-          numSubTypeCount = calcTypeCount(cn.value.filter { cnc -> cnc.caseNoteSubType == subType })
```
Mjwillis, line 155: Aren't interested in the sub type count anymore

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/BehaviourService.kt
```diff
@@ -0,0 +1,69 @@
...
+      }.associateBy { it.key }
+
+  private fun calcTypeCount(caseNoteUsage: List<PrisonerCaseNoteByTypeSubType>): Int =
+    caseNoteUsage.map { it.numCaseNotes }.fold(0) { acc, next -> acc + next }
+
+  private fun getLastRealReviewForOffenders(reviews: List<PrisonerIepLevel>): Map<Long, LocalDateTime> =
+    reviews
+      .groupBy { it.bookingId }
+      .map { review ->
+        val latestReview = review.value.firstOrNull(PrisonerIepLevel::isRealReview) ?: review.value.first()
```
xoen, line 45: Are we sure this `?: review.value.first()` is the correct thing to do?

Someone may have been in prison for months/years without a genuine Incentive review.
They may have a number of relevant case notes in the last 3 months.
But if, let's say, they got transferred last week this `review.value.first()` would mean only case notes from the last week would returned, right?

Maybe this should fallback to 3 months ago when no real review is found? Sorry if this is a stupid question, I'm still working my way through this PR.

Mjwillis: Its a good point. I could change it so that if no real review found then just get three months

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsService.kt
```diff
@@ -63,11 +63,8 @@ class IncentiveReviewsService(
     val offenders = deferredOffenders.await()
 
     val bookingIds = offenders.map(OffenderSearchPrisoner::bookingId)
-    val prisonerNumbers = offenders.map(OffenderSearchPrisoner::prisonerNumber)
-
-    val deferredPositiveCaseNotesInLast3Months = async { getCaseNoteUsage("POS", "IEP_ENC", prisonerNumbers) }
-    val deferredNegativeCaseNotesInLast3Months = async { getCaseNoteUsage("NEG", "IEP_WARN", prisonerNumbers) }
 
+    val deferredBehaviourCaseNotesSinceLastReview = async { behaviourService.getBehaviours(prisonerIepLevelRepository.findAllByBookingIdInOrderByReviewTimeDesc(bookingIds = bookingIds).toList()) }
```
xoen, line 67: I'm not sure how big is the window of opportunity here but potentially someone may have no entries and therefore not be in this list (even tho they may have a some case notes maybe?)

xoen: Another thing to bear in mind is that this list could be very very big, especially if the wing has a few prisoners which have been in prison for years and years and have a lot of records.

In theory we're only interested in the last 3 months, but of course adding that to the query means you would remove prisoners without a recent review from the list, and therefore their case notes would be returned.

Mjwillis: Every prisoner MUST have an incentive review, in which case they will all be considered -Good point though!

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/BehaviourService.kt
Mjwillis: @xoen I've updated this to your suggestion

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
```diff
@@ -41,21 +41,22 @@ class IncentiveSummaryService(
...
     val provenAdjudications = async { getProvenAdjudications(bookingIds) }
-    val iepDetails = getIEPDetails(bookingIds)
 
-    val positiveCount = positiveCaseNotes.await()
-    val negativeCount = negativeCaseNotes.await()
+    val reviews = prisonerIepLevelRepository.findAllByBookingIdInOrderByReviewTimeDesc(bookingIds = bookingIds).toList()
+    val iepDetails = getCurrentAndHistoricalReviews(reviews).associateBy(IepResult::bookingId)
+
+    val deferredBehaviourCaseNotesSinceLastReview = async {
+      behaviourService.getBehaviours(prisonerIepLevelRepository.findAllByBookingIdInOrderByReviewTimeDesc(bookingIds = bookingIds).toList())
```
xoen, line 51: I may miss something but you could reuse the `reviews` value here AFAICT should contain the same list so we can save the trip to the DB.

xoen (commented):
LGTM, however this is probably worth changing before merging :)

https://github.com/ministryofjustice/hmpps-incentives-api/pull/360/files/1baec61de94aed920a2dadb3c9bf22ae8e94a298#r1109960720

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveSummaryService.kt
Mjwillis: Duh, I specifically split this out so I could do that, must have been a cut and paste job when refactoring!  Will change

Mjwillis:
@xoen update PR based on your very good suggestions!

xoen (approved):

PR #379, opened 2023-03-16
INC-1119: use new HMPPS Domain Events queue from the Incentives namespace - Prod

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-1119


helm_deploy/hmpps-incentives-api/values.yaml
```diff
@@ -32,6 +32,40 @@ generic-service:
     APPLICATIONINSIGHTS_CONNECTION_STRING: "InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY)"
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
 
+  # Pre-existing kubernetes secrets to load as environment variables in the deployment.
```
neilmendum, line 35: Moving this back here, which reverts https://github.com/ministryofjustice/hmpps-incentives-api/pull/368, as all environments should use the following secrets
- `sqs-incentives-secret` ->  `sqs-prisoner-event-queue-secret`
- `sqs-incentives-dlq-secret` -> `sqs-prisoner-event-queue-dlq-secret`

ushkarev (approved):

PR #381, opened 2023-03-16
INC-1142 & INC-1143 Part 1: Add endpoints to manage (global) incentive levels

ushkarev:
NB:
• the backing data is _not_ yet used (deferring to prison-api for global incentive level configuration from NOMIS)
• the backing data is _not_ yet synchronised to NOMIS
• endpoints that permit modification will require a special role but this is not yet set up
• the repercussions of disabling an incentive level globally are significant, but at this point, there are no restrictions


ushkarev:
☝️ I made sure each commit passed tests (ignoring sonar) and recommend reviewing commit-by-commit because many lines change (mostly boilerplate)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/HmppsIncentivesApiExceptionHandler.kt
```diff
@@ -69,7 +71,7 @@ class HmppsIncentivesApiExceptionHandler {
   }
 
   @ExceptionHandler(ParameterValidationException::class)
-  fun handleValidationException(e: ParameterValidationException): ResponseEntity<ErrorResponse> {
+  fun handleParameterValidationException(e: ParameterValidationException): ResponseEntity<ErrorResponse> {
```
ushkarev, line 74: It might not have been a problem due to annotation, but this method was not uniquely names

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveSummaryMasteredResourceTest.kt
```diff
@@ -136,15 +140,15 @@ class IncentiveSummaryMasteredResourceTest : SqsIntegrationTestBase() {
...
+  private suspend fun prisonerIepLevel(
     bookingId: Long,
     prisonerNumber: String,
     level: String = "STD",
     reviewType: ReviewType = ReviewType.REVIEW,
     reviewTime: LocalDateTime = now(),
     current: Boolean = false,
     prisonId: String = "MDI"
-  ) = repository.save(
+  ) = prisonerIepLevelRepository.save(
```
ushkarev, line 151: These entities used to remain after test case completion, restricting incentive level tests.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/IncentiveLevelRepository.kt
```diff
@@ -0,0 +1,12 @@
+package uk.gov.justice.digital.hmpps.incentivesapi.jpa.repository
+
+import kotlinx.coroutines.flow.Flow
+import org.springframework.data.repository.kotlin.CoroutineCrudRepository
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.incentivesapi.jpa.IncentiveLevel
+
+@Repository
+interface IncentiveLevelRepository : CoroutineCrudRepository<IncentiveLevel, String> {
+  fun findAllByOrderBySequence(): Flow<IncentiveLevel>
```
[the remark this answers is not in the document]
ushkarev: it's the r2dbc convention
https://docs.spring.io/spring-data/r2dbc/docs/current/reference/html/#appendix.query.method.subject
…after all, it has to inspect the function name to decide what it means

src/main/resources/db/migration/V1_25__incentive_levels.sql
```diff
@@ -0,0 +1,25 @@
...
+CREATE INDEX incentive_level_sequence_idx ON incentive_level ("sequence");
+CREATE INDEX incentive_level_active_idx ON incentive_level (active);
+
+INSERT INTO incentive_level (code, description, "sequence", active)
+VALUES ('BAS', 'Basic', 1, true),
+       ('STD', 'Standard', 2, true),
+       ('ENH', 'Enhanced', 3, true),
+       ('EN2', 'Enhanced 2', 4, true),
+       ('EN3', 'Enhanced 3', 5, true),
+       ('ENT', 'Entry', 99, false);
```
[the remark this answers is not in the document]
ushkarev: This included set is exactly what's in NOMIS

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/HmppsIncentivesApiExceptionHandler.kt
```diff
@@ -231,6 +245,9 @@ class HmppsIncentivesApiExceptionHandler {
 class NoDataFoundException(id: Long) :
   Exception("No Data found for ID $id")
 
+class NoDataWithCodeFoundException(dataType: String, code: String) :
+  Exception("No $dataType found for code $code")
```
xoen, line 271: I tend to surround these kind of strings in single quote in case there is some accidental whitespace, that can help, e.g. `No IncentiveLevel found for code 'STD '` is less puzzling in case of strange 404s :) (probably more useful when making manual requests to API than in machine-to-machine requests tho)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/IncentiveLevelRepository.kt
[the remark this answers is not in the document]
xoen: Ah so the first `By` is necessary after all, I thought it would have been clever enough :)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveLevelService.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.incentivesapi.jpa.IncentiveLevel
+import uk.gov.justice.digital.hmpps.incentivesapi.jpa.repository.IncentiveLevelRepository
+import uk.gov.justice.digital.hmpps.incentivesapi.dto.IncentiveLevel as IncentiveLevelDTO
+
+@Service
+class IncentiveLevelService(
+  private val incentiveLevelRepository: IncentiveLevelRepository,
+) {
+  suspend fun getIncentiveLevels(): List<IncentiveLevelDTO> {
```
xoen, line 23: I would personally add the word `All` to this method.
To make it more explicit it's returning all levels including non-active (without having to look at its definition), e.g. `getAllIncentiveLevels()`.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveLevelResource.kt
```diff
@@ -86,4 +92,133 @@ class IncentiveLevelResource(
...
+    @Schema(description = "Incentive level code", example = "STD", required = true, minLength = 3, maxLength = 6)
+    @PathVariable code: String,
+    @RequestBody update: IncentiveLevelUpdate
+  ): IncentiveLevel {
+    ensure {
+      update.description?.let {
+        ("description" to it).hasLengthAtLeast(1)
+      }
+    }
+    return incentiveLevelService.updateIncentiveLevel(code, update)
```
xoen, line 262: Same comment as above, the implementation suggests maybe the code for the PUT endpoint got mixed up with the code for the PATCH endpoint (verb traditionally used for partial updates)

Also, I notice both methods get the same DTO class, how do these endpoint distinguish the two?

TBH I'm wondering if we could just have the PATCH endpoint (more flexible) - do we need the PUT endpoint too? (genuine question, I may miss something)

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/IncentiveLevelRepositoryTest.kt
```diff
@@ -0,0 +1,62 @@
...
+@ActiveProfiles("test")
+@WithMockUser
+class IncentiveLevelRepositoryTest : TestBase() {
+  @Autowired
+  lateinit var repository: IncentiveLevelRepository
+
+  @Test
+  fun `incentive levels properly set up`(): Unit = runBlocking {
+    repository.findAllByOrderBySequence().map { it.sequence }.reduce { previous, sequence ->
+      assertThat(sequence).isGreaterThan(previous)
```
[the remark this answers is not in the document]
ushkarev: This is deliberate! I don't want sequence numbers to be the same as then ordering won't be stable. The api does not expose the sequence number so as long as initial data is fine, then sequence numbers are not going to clash.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveLevelResource.kt
```diff
@@ -86,4 +92,133 @@ class IncentiveLevelResource(
...
+    @Schema(description = "Incentive level code", example = "STD", required = true, minLength = 3, maxLength = 6)
+    @PathVariable code: String,
+    @RequestBody incentiveLevel: IncentiveLevel
+  ): IncentiveLevel {
+    ensure {
+      if (code != incentiveLevel.code) {
+        errors.add("Incentive level codes must match in URL and payload")
+      }
+    }
+    return partiallyUpdateIncentiveLevel(code, incentiveLevel.toUpdate())
```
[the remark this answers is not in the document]
ushkarev: The PATCH endpoint accepts a superset of payloads with respect to the PUT endpoint, so it's better to call `partiallyUpdateIncentiveLevel` from `updateIncentiveLevel` than repeat the same code.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveLevelResource.kt
ushkarev: The two endpoints accept different request bodies! PUT requires a complete and valid object while PATCH accepts a partial one.

On the point of do we need it: to be honest, we don't really need any of the modifying endpoints as it'll essentially never happen ;)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveLevelService.kt
ushkarev: ✅

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/HmppsIncentivesApiExceptionHandler.kt
ushkarev: ✅

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveLevelResource.kt
ushkarev: I agree it's overkill, but 🤷

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveLevelResource.kt
```diff
@@ -99,6 +99,46 @@ class IncentiveLevelResource(
     return incentiveLevelService.createIncentiveLevel(incentiveLevel)
   }
 
+  @PatchMapping("level-order")
```
xoen, line 102: It may work better in plural, e.g. `PATCH /incentive/levels-order`?

xoen (approved):

neilmendum (approved):
LGTM

PR #383, opened 2023-03-20
INC-1142 & INC-1143 & INC-1145 Part 2: Add endpoints to manage prison incentive level information

ushkarev:
NB:
• the backing data is not yet used (deferring to prison-api for global & per-prison incentive level configuration from NOMIS)
• the backing data is not yet synchronised to NOMIS
• endpoints that permit modification will require a special role but this is not yet set up
• the repercussions of disabling an incentive level globally are significant, but at this point, there are no restrictions
• the repercussions of disabling an incentive level in a prison are not insignificant, but at this point, there are few restrictions

Cf. #381 for part 1


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveLevelResource.kt
```diff
@@ -266,16 +270,16 @@ class IncentiveLevelResource(
...
+    summary = "Deactivates an incentive level",
+    // TODO: decide and explain what happens to associated per-prison data
     responses = [
       ApiResponse(
         responseCode = "200",
-        description = "Incentive level disabled"
+        description = "Incentive level deactivated"
       ),
       // ApiResponse(
       //   responseCode = "400",
```
Mjwillis, line 281: Is this meant to be commented?

Mjwillis (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveLevelResource.kt
ushkarev: yes, i've deliberately left lots of TODOs as discussion points! once all the business rules have been determined, then can get rid of all of these

xoen:
FYI: I know this is already merged but I'm planning to "review" this later today.

PR #386, opened 2023-03-21
INC-1143: Use sensible spend limit defaults from PSI

ushkarev:
This does not add any validation/ranges to spending limits; simply sets the fallback values for any newly-created incentive level information for a prison in cases where an incomplete payload is submitted.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepositoryTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepositoryTest.kt
@@ -65,10 +65,10 @@ class PrisonIncentiveLevelRepositoryTest : TestBase() {
       levelCode = "STD",
       prisonId = "MDI",
 
-      remandTransferLimitInPence = 2500,
-      remandSpendLimitInPence = 25000,
-      convictedTransferLimitInPence = 500,
-      convictedSpendLimitInPence = 5000,
+      remandTransferLimitInPence = 2750,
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepositoryTest.kt
```diff
@@ -65,10 +65,10 @@ class PrisonIncentiveLevelRepositoryTest : TestBase() {
       levelCode = "STD",
       prisonId = "MDI",
 
-      remandTransferLimitInPence = 2500,
-      remandSpendLimitInPence = 25000,
-      convictedTransferLimitInPence = 500,
-      convictedSpendLimitInPence = 5000,
+      remandTransferLimitInPence = 2750,
```
xoen, line 68: Any reason why you didn't use the `_` to delimit the pence in the tests as well like in the `spendLimitPolicy` value? I quite like it and it helps readability.

xoen: Maybe to avoid possible confusion with the values in the JSON strings?

xoen (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepositoryTest.kt
ushkarev: no good reason :) will replace in next pr

PR #389, opened 2023-03-22
INC-1148: Audit changes to incentive levels and their per-prison information

ushkarev:
Submit SQS messages when an `IncentiveLevel` or `PrisonIncentiveLevel` is modified.

NB: For now, this happens inside transactions so data will not be committed to database at the point the message is received somewhere. This shouldn't hugely matter in case of audit events, but "HMPPS domain events" will need to be outside transactions so suggest moving the audit submissions too at that point.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveLevelService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveLevelService.kt
@@ -111,7 +111,8 @@ class IncentiveLevelService(
     return incentiveLevelRepository.saveAll(incentiveLevelsWithNewSequences)
       .toListOfDTO()
       .also {
-        auditService.sendMessage(INCENTIVE_LEVELS_REORDERED, "*", it.map(IncentiveLevelDTO::code))
+        val orderedCodes = it.map(IncentiveLevelDTO::code)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveLevelService.kt
```diff
@@ -97,7 +108,11 @@ class IncentiveLevelService(
       )
     }
 
-    return incentiveLevelRepository.saveAll(incentiveLevelsWithNewSequences).toListOfDTO()
+    return incentiveLevelRepository.saveAll(incentiveLevelsWithNewSequences)
+      .toListOfDTO()
+      .also {
+        auditService.sendMessage(INCENTIVE_LEVELS_REORDERED, "*", it.map(IncentiveLevelDTO::code))
```
xoen, line 114: I'm not sure how the ID is used on the other side of the fence. I'm guessing it's not, just stored for reference/displayed somewhere.

However, even if that's so, I'd probably prefer something more specific here.

I know this is an operation which didn't involve a single entity but what do you think about having the list of incentive codes here instead? e.g. `"BAS, STD, ENH, EN2, EN3"` - it's an idea.

xoen (approved):

PR #391, opened 2023-03-24
INC-1171: Allow globally-defined incentive levels to require per-prison information to be active

ushkarev:
New business rules:
• If an incentive level is globally required, per-prison levels and associated information cannot be deactivated
• If an incentive level becomes globally required, it will be activated automatically in all prisons that currently have some active levels (essentially, in all active prisons); associated information will use policy defaults

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveLevelResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveLevelResource.kt
@@ -185,8 +185,9 @@ class IncentiveLevelResource(
   @PreAuthorize("hasRole('MAINTAIN_INCENTIVE_LEVELS') and hasAuthority('SCOPE_write')")
   @Operation(
     summary = "Updates an incentive level",
-    description = "Payload must include all required fields",
-    // TODO: decide and explain what happens to associated per-prison data
+    description = "Payload must include all required fields. A level marked as required must also be active.",
+    // TODO: decide and explain what happens to associated per-prison data when activating/deactivating;
+    //       especially if level is active and/or occupied in some prison
     responses = [
       ApiResponse(
         responseCode = "200",
@@ -194,7 +195,7 @@ class IncentiveLevelResource(
       ),
       ApiResponse(
         responseCode = "400",
-        description = "Invalid payload", // TODO: maybe also when deactivating and there are prisons with level activated?
+        description = "Invalid payload",
         content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
       ),
       ApiResponse(
@@ -233,8 +234,9 @@ class IncentiveLevelResource(
   @PreAuthorize("hasRole('MAINTAIN_INCENTIVE_LEVELS') and hasAuthority('SCOPE_write')")
   @Operation(
     summary = "Updates an incentive level",
-    description = "Partial updates are allowed",
-    // TODO: decide and explain what happens to associated per-prison data
+    description = "Partial updates are allowed. A level marked as required must also be active.",
+    // TODO: decide and explain what happens to associated per-prison data when activating/deactivating;
+    //       especially if level is active and/or occupied in some prison
     responses = [
       ApiResponse(
         responseCode = "200",
@@ -242,7 +244,7 @@ class IncentiveLevelResource(
       ),
       ApiResponse(
         responseCode = "400",
-        description = "Invalid payload", // TODO: maybe also when deactivating and there are prisons with level activated?
+        description = "Invalid payload",
         content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
       ),
       ApiResponse(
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveLevelResource.kt
```diff
@@ -311,6 +323,6 @@ class IncentiveLevelResource(
     @PathVariable
     code: String,
   ): IncentiveLevel {
-    return partiallyUpdateIncentiveLevel(code, IncentiveLevelUpdate(active = false))
```
xoen, line 314: We're not checking for `required` here to validate the request. We could.

I'm guessing we're trusting the user to know what they're doing and avoiding doing this in 2 steps?:
- mark a level as optional
- then deactivate a level

ushkarev: i guess it needs better explanation in the swagger docs at least.
we still don't really know what happens when a level is deactivated globally (as yet, nothing). 🤷

xoen: True.

Part of me thinks this is simpler but the other part wonders whether not having the validation means it's easy to accidentally deactivate a _required_ level (as it would set `required=false` behind the scenes for the user/client)

ushkarev: well the user here is just our team cos i don't see why anybody else would want to delete levels ;)
i'll look at the swagger docs at least, but can't decide whether patch + delete is actually better. i mean, the delete action is really dangerous irrespective of whether a level is required.

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepositoryTest.kt
```diff
@@ -240,4 +241,34 @@ class PrisonIncentiveLevelRepositoryTest : TestBase() {
...
+          remandTransferLimitInPence = 6050,
+          remandSpendLimitInPence = 60500,
+          convictedTransferLimitInPence = 1980,
+          convictedSpendLimitInPence = 19800,
+
+          visitOrders = 2,
+          privilegedVisitOrders = 1,
+
+          new = true,
+        )
```
xoen, line 266: You could your fancy `makePrisonIncentiveLevel()` helper and then set `active=false` and that would be more intent revealing/simpler (assuming I'm getting what you're doing here right, e.g. you're setting all levels in EXI/LEI prisons to inactive)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveLevelService.kt
```diff
@@ -67,6 +71,10 @@ class IncentiveLevelService(
     return incentiveLevelRepository.save(incentiveLevel)
       .toDTO()
       .also {
+        if (it.required) {
+          // This probably could be moved out of the transaction once auditing is moved out:
```
xoen, line 75: Assuming enabling the level in all prisons is quick enough it may make sense to keep this within the same transaction.

The two operations kinda make sense only together no? 🤔

ushkarev: well i'm pretty sure it would be up to 120 _separate_ DB updates cos the values come from this service and the `repository.saveAll(…)` returns a flow – that implies it updates one-by-one. i don't mind the prisons picking up the change "a little later". still, right now it's inside the transaction :)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveLevelService.kt
```diff
@@ -73,15 +85,36 @@ class IncentiveLevelService(
...
+            if (it.required && !originalIncentiveLevel.required) {
+              // This probably could be moved out of the transaction once auditing is moved out:
+              enablePrisonIncentiveLevelEverywhere(it.code)
+            }
             auditService.sendMessage(INCENTIVE_LEVEL_UPDATED, it.code, it)
           }
       }
   }
 
+  // TODO: Many DB queries are made: does this warrant NOT being in a transaction?
```
Mjwillis, line 107: Won't this be in a transaction anyway as its private and I assume `@Transactional` is already setup higher up

Mjwillis (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveLevelService.kt
ushkarev: the methods that call this are transactional currently, my todo comments suggests maybe moving this out of the transaction. here's some more discussion on that point:

https://github.com/ministryofjustice/hmpps-incentives-api/pull/391#discussion_r1149167238

xoen (commented):
FYI: This looks great so far 👍 - But I know you're adding more tweaks to this PR so I'll have a look at the rest when they're in ☺️

xoen (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepositoryTest.kt
```diff
@@ -190,26 +190,27 @@ class PrisonIncentiveLevelRepositoryTest : TestBase() {
...
+  private fun makeNewEntity(levelCode: String, prisonId: String) = PrisonIncentiveLevel(
+    levelCode = levelCode,
+    prisonId = prisonId,
+    active = levelCode != "ENT",
+    defaultOnAdmission = levelCode == "STD",
+
+    remandTransferLimitInPence = 6050,
+    remandSpendLimitInPence = 60500,
+    convictedTransferLimitInPence = 1980,
+    convictedSpendLimitInPence = 19800,
```
xoen, line 202: You could use the `_` as pence separator here for consistency if you'd like

PR #394, opened 2023-03-29
INC-1164: Post National rollout cleanup

xoen:
### Important  
⚠️ Do not merge this PR until we're sure the old table is not used anymore.

### Context
On Monday we'll rollout the new 'Reviews' table to all prisons (see [related PR](https://github.com/ministryofjustice/hmpps-incentives-ui/pull/435))

As a result, we can remove a bunch of classes and code which will soon be unused 🎊 

Similarly, it has been months now since we moved away from getting Incentives records from NOMIS/Prison API. The transition period is over, which means we can simplify `IepSummary` and `IepDetail` data classes to have less optional fields.
Some of their fields like `id` or `reviewType` were optional because an incentive record data could either come from our DB or NOMIS/Prison API - were these wouldn't be present. However now all Incentives records data is coming from our DB and we know what's mandatory and what's optional.

### Summary of changes
- The main change is the removal of the deprecated`GET /incentives-summary/prison/{prisonId}/location/{locationId}` endpoint only used by the Incentives API and corresponding resource class
  - On Monday 3rd April all prisons will be switched to use the new 'Reviews' table which uses a separate endpoint (`GET /incentives-reviews/prison/{prisonId}/location/{cellLocationPrefix}/level/{levelCode}`)
  - as a result the related `IncentiveSummaryService` class only used by the old endpoint is gone
  - a bunch of classes like `BehaviourSummary`/`IncentiveLevelSummary`/`PrisonerIncentiveSummary` used for the response of this endpoint are also gone 
  - support classe like `NoPrisonersAtLocationException`, `IepResult` and `SortColumn` only used in this method are also gone
  - proven adjudications are not retrieved anymore so I've removed the corresponding service method and data classes
- The other big change is making more fields non-nullable
  - Incentive reviews records are now taken from our database and the transition away from NOMIS/Prison API is complete (for Incentive reviews at least)
  - During this transition period data could have come from NOMIS/Prison API where some fields were not available, and hence optional in the DTO classes
  - A good example of this is stuff like `id`, `prisonerNumber` and `reviewType` on the `IepDetail`/`IepSummary` classes
    - `id` was nullable because it's the value for the primary key in our database, was not available for data coming from Prison API
    - `prisonerNumber` is a DPS concept which was not available in the data coming from NOMIS/Prison API
    - `reviewType` is something we introduced with the new Incentives service to discriminate between automated Incentive entries (e.g. for admission or transfer) and genuine Incentive reviews. This is something data coming from NOMIS/Prison API didn't have
    - as a result these fields can now be marked as mandatory in the DTO classes meaning the code dealing with those can be less defensive (e.g. I was able to remove some `iepDetail.id?.let {}` blocks)
    - all the fields above are always present in all rows in `prisoner_iep_level` DB table. Furthermore` reviewType` which is always mandatory was for NULL-able at DB level so I've added a small migration to be sure it can't be NULL (as expected)
    - similarly `auditModuleName` is not persisted but it's always set to `"INCENTIVES_API"` (value in `SYSTEM_USERNAME` constant) throughout the codebase so I've made that mandatory too
    - The `IsRealReview` interface could be simplified as it doesn't have to check for `null` review type anymore (we treated these reviews as "real" as they were coming from NOMIS/Prison API and there was no way to determine whether they were genuine Incentives reviews)
    - `nextReviewDate` in `IepSummary` was always set through the codebase so I've made it mandatory too. This was optional in the past because the data could have come from Prison API without the historic list of Incentive reviews necessary to calculate this field. This is no longer the case.
  

### How to review
You should be able to review this commit-by-commit. Commit messages also include additional context.


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
```diff
@@ -14,14 +14,14 @@ import java.time.LocalDateTime
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @Schema(description = "IEP Review Summary for Prisoner")
 data class IepSummary(
-  @Schema(description = "Unique ID for this review (new Incentives data model only)", required = false, example = "12345")
-  val id: Long? = null,
+  @Schema(description = "Unique ID for this review (new Incentives data model only)", required = true, example = "12345")
+  val id: Long,
```
ushkarev, line 18: <3

Mjwillis:
Lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/OffenderSearchService.kt
```diff
@@ -34,14 +34,4 @@ class OffenderSearchService(private val offenderSearchWebClient: WebClient) {
     } while (!pageOfData.last)
     return offenders
   }
-
-  /**
-   * Gets one offender record
-   * Requires ROLE_PRISONER_SEARCH or ROLE_VIEW_PRISONER_DATA role
-   */
-  suspend fun getOffender(prisonerNumber: String) =
```
ushkarev, line 42: meanie. getting rid of my code that was never used 😜

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IepSummary.kt
```diff
@@ -76,9 +76,9 @@ data class IepDetail(
   @Schema(description = "Username of the reviewer", required = true, example = "USER_1_GEN")
   val userId: String?,
   @Schema(description = "Type of IEP Level change", required = true, example = "REVIEW")
-  override val reviewType: ReviewType? = null,
+  override val reviewType: ReviewType,
   @Schema(description = "Internal audit field holding which system/screen recorded the review", required = true, example = SYSTEM_USERNAME)
-  val auditModuleName: String? = null,
+  val auditModuleName: String = SYSTEM_USERNAME,
```
ushkarev, line 81: convenient

src/main/resources/db/migration/V1_29__prisoner_iep_level_make_review_type_non_nullable.sql
```diff
@@ -0,0 +1,3 @@
+-- Make review_type mandatory
+ALTER TABLE prisoner_iep_level
+ALTER COLUMN review_type SET NOT NULL;
```
ushkarev, line 3: still valid for `prod` right now

Mjwillis (approved):
Just need to resolve merge issues

ushkarev (approved):

PR #396, opened 2023-03-29
INC-1141: Allow retrieving inactive per-prison incentive levels

ushkarev:
This makes the endpoints more symmetrical with respect to global incentive levels and accommodates the need to synchronise data into other services _when it is deactivated_


doc/architecture/decisions/0004-reference-data-sync.md
```diff
@@ -270,8 +270,20 @@ Authorised requests do not require any roles.
 ]
 ```
 
+#### Get details of an incentive level globally
+NB: Inactive levels are also returned
```
neilmendum, line 274: I think this is sensible as the [PrisonAPI endpoint we call](https://github.com/ministryofjustice/hmpps-incentives-api/blob/66f95d189c6416a9c1de7d1d331e99612586f0bb/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonApiService.kt#L57) returns active & inactive levels

neilmendum (approved):
LGTM

doc/architecture/decisions/0004-reference-data-sync.md
ushkarev: Perhaps I was going too far with tidying up the external interface :)

doc/architecture/decisions/0004-reference-data-sync.md
```diff
@@ -296,8 +308,14 @@ e.g. **/incentive/prison-levels/MDI** returns
 ]
 ```
 
+#### Get the details of all active incentive levels for a specified prison
+This response is the same as above but includes `"active": false`
+
+`GET /incentive/prison-levels/{prisondId}?with-inactive=true`
```
xoen, line 314: I like that by default only active are returned but it's possible to request all of them when needed 👍

ushkarev: i would have preferred to hide inactive ones but then we cannot tell nomis something has been deactivated :(

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonIncentiveLevelService.kt
```diff
@@ -32,6 +32,13 @@ class PrisonIncentiveLevelService(
   private val incentiveLevelRepository: IncentiveLevelRepository,
   private val prisonIncentiveLevelRepository: PrisonIncentiveLevelRepository,
 ) {
+  /**
+   * Returns all incentive levels for given prison, including inactive ones, along with associated information, in globally-defined order
+   */
+  suspend fun getAllPrisonIncentiveLevels(prisonId: String): List<PrisonIncentiveLevelDTO> {
+    return prisonIncentiveLevelRepository.findAllByPrisonId(prisonId).toListOfDTO()
```
xoen, line 39: It's an interesting one at prison level 🤔 

If a `(prisonId, levelCode)` row is not present for whatever reason it wouldn't be in the list/returned.

This will only include levels which were at some point added to a prison, including the ones that were deactivated. 
This could be a bit unpredictable.

I think it makes a lot of sense for the endpoint returning the "global" list of Incentive levels to optionally return all of the levels.

However what's the use case for the prison-level list endpoint to return "all" levels?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/PrisonIncentiveLevelResource.kt
```diff
@@ -54,13 +55,21 @@ class PrisonIncentiveLevelResource(
     @Schema(description = "Prison id", example = "MDI", required = true, minLength = 3, maxLength = 6)
     @PathVariable
     prisonId: String,
+    @Schema(description = "Include inactive prison incentive levels", example = "true", required = false, defaultValue = "false", type = "boolean", pattern = "^[true|false]$")
+    @RequestParam(defaultValue = "false", value = "with-inactive", required = false)
+    withInactive: Boolean = false,
   ): List<PrisonIncentiveLevel> {
-    return prisonIncentiveLevelService.getActivePrisonIncentiveLevels(prisonId)
+    return if (withInactive) {
+      prisonIncentiveLevelService.getAllPrisonIncentiveLevels(prisonId)
```
xoen, line 63: This would only include the levels for which there is a `(prisonId, levelCode)` row in the DB tho right?

There is no guarantee this list would be exhaustive and include all levels available/active across all the prison estate right?

ushkarev: correct. but i didn't want to force-create levels in a prison using default values for something historic cos the defaults are unlikely to make sense… but to the consumer of the api would look like historical truth!

ushkarev: this endpoint's documentation will now better explain the situation #398

PR #398, opened 2023-03-30
INC-1188: Test domain events are raised when incentive levels are modified

ushkarev:
When incentive levels or per-prison associated data is modified, HMPPS domain events are sent using SNS. Now tests assert that this happens at appropriately. This ensures that changes in #395 are correctly implemented.

Also might as well document the api endpoints more thoroughly.


neilmendum (approved):
LGTM

PR #400, opened 2023-04-01
INC-1190: Refactor incentive level services so that audited forms are subclasses

ushkarev:
This is a refactor of auditing introduced in #395. The point is to ensure that only audited methods are used in resources (because the resource does not have access to an unaudited service instance).

## Before any audit services existed

```mermaid
classDiagram
    IncentiveLevelService <-- IncentiveLevelResource : uses
    PrisonIncentiveLevelService <-- PrisonIncentiveLevelResource : uses
```

## When 1 spanning audit service was created (the starting point for this PR)
[Diff introducing auditing](https://github.com/ministryofjustice/hmpps-incentives-api/compare/43c6a591ef3a2fc61c4e5d5f4ee9e8dfa15a8a45..66f95d189c6416a9c1de7d1d331e99612586f0bb)

```mermaid
classDiagram
    IncentiveLevelService <-- IncentiveLevelEventAuditCurator : uses (modifying)
    PrisonIncentiveLevelService <-- IncentiveLevelEventAuditCurator : uses (modifying)
    IncentiveLevelEventAuditCurator <-- IncentiveLevelResource : uses (modifying)
    IncentiveLevelService <-- IncentiveLevelResource : uses (read-only)
    IncentiveLevelEventAuditCurator <-- PrisonIncentiveLevelResource : uses (modifying)
    PrisonIncentiveLevelService <-- PrisonIncentiveLevelResource : uses (read-only)
```

## After this PR
[Eventual diff with separate auditing](https://github.com/ministryofjustice/hmpps-incentives-api/compare/43c6a591ef3a2fc61c4e5d5f4ee9e8dfa15a8a45..audited-incentive-level-services) – note that resources only use audited services

```mermaid
classDiagram
    IncentiveLevelService <|-- IncentiveLevelAuditedService : extends
    PrisonIncentiveLevelService <|-- PrisonIncentiveLevelAuditedService : extends

    IncentiveLevelAuditedService <-- IncentiveLevelResource : uses
    PrisonIncentiveLevelAuditedService <-- PrisonIncentiveLevelResource : uses
```


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveLevelAuditedService.kt
```diff
@@ -0,0 +1,73 @@
...
+class IncentiveLevelAuditedService(
+  private val clock: Clock,
+  private val auditService: AuditService,
+  private val snsService: SnsService,
+  incentiveLevelRepository: IncentiveLevelRepository,
+  prisonIncentiveLevelAuditedService: PrisonIncentiveLevelAuditedService,
+) : IncentiveLevelService(
+  clock = clock,
+  incentiveLevelRepository = incentiveLevelRepository,
+  prisonIncentiveLevelService = prisonIncentiveLevelAuditedService,
```
ushkarev, line 25: Note that _audited_ service is passed to superclass

PR #402, opened 2023-04-04
`NextReviewDateService` no longer uses `IepDetail`

xoen:
The `IepDetail` class is still used, and in fact even part of the public interface
of some of the endpoints. However is kinda a legacy data class.

The `NextReviewDateService` now uses `PrisonerIepLevel` values which is
what's stored in the database. This is the new representation of an incentive
record.


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/PrisonerIepLevel.kt
```diff
@@ -34,4 +36,8 @@ data class PrisonerIepLevel(
   override fun isNew(): Boolean = new
 
   override fun getId(): Long = id
+
+  @Transient
+  @ReadOnlyProperty
+  val reviewDate: LocalDate = reviewTime.toLocalDate()
```
xoen, line 42: This is a convenience property for when code only cares about the date not the datetime.

It's to avoid having to call `.reviewTime.toLocalDate()` all the times.

ushkarev (approved):
looks legit ;)

PR #403, opened 2023-04-04
ADR: Rename Incentive level "description" to "name"

xoen:
Reason: The current values for "description", e.g. "Standard" or "Enhanced" are the **names** of the Incentive levels, NOT their description.

The new Incentive levels management UI will show levels descriptions describing the level so it makes sense to rename the existing "description" field.
The new "description" fields will be non-Nullable but default to an empty string `""`.

**NOTE**: Prison Incentive levels have a field called `levelDescription`, this has the same meaning and will be renamed as well to `levelName` (a new `levelDescription` field will be added but it will have a different meaning)

### DIFF
--- a/doc/architecture/decisions/0004-reference-data-sync.md
+++ b/doc/architecture/decisions/0004-reference-data-sync.md
@@ -255,13 +255,15 @@ Authorised requests do not require any roles.
 [
   {
     "code": "BAS",
-    "description": "Basic",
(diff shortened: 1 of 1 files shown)


xoen:
This ADR update is included [in this other PR (merged)](https://github.com/ministryofjustice/hmpps-incentives-api/pull/405) which supersede this one. Closing this PR.

PR #404, opened 2023-04-04
INC-1147: Generically typed domain event

ushkarev:
HMPPS domain events have a nullable field (`additionalInformation`) that can contain any old object, it depends on the event type. Right now, we model this field using a highly polymorphic class with nullable fields for every possible use. This prevents us from using strong typing to ensure correct fields are used in the correct places.

This PR is an attempt to make the domain event generic over this field.

I'm not totally convinced by this! Generics + Jackson deserialisation do not always play well together… but tests appear to pass so assuming our coverage is good enough then…?


src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/integration/IncentiveLevelResourceTestBase.kt
```diff
@@ -76,27 +78,31 @@ class IncentiveLevelResourceTestBase : SqsIntegrationTestBase() {
...
     assertThat(attributes.eventType.Value).isEqualTo(eventType)
-    val domainEvent = objectMapper.readValue(message, HMPPSDomainEvent::class.java)
+    val domainEvent = objectMapper.readValue(message, typeReference)
     assertThat(domainEvent.eventType).isEqualTo(eventType)
 
     return domainEvent
   }
 
-  protected fun getPublishedDomainEvents(): List<HMPPSDomainEvent> {
+  protected fun getPublishedDomainEvents(): List<HMPPSDomainEvent<Map<String, *>>> {
```
ushkarev, line 99: `Map<String, *>` is an effective super set of `IncentiveLevelRef` and `PrisonIncentiveLevelRef`

PR #409, opened 2023-04-06
INC-1197: Remove migration and sync endpoints

xoen:
⚠️ **NOTE**: Do not merge this PR until Syscon confirmed they removed the code using these endpoints on their end.

You can review this PR commit by commit. Commit messages should provide some context about the specific change.

TODO: Write a PR description to explain what's going on


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveStoreService.kt
```diff
@@ -40,19 +40,6 @@ class IncentiveStoreService(
     nextReviewDateUpdaterService.update(remainingBookingId)
   }
 
-  suspend fun deleteIncentiveReview(prisonerIepLevel: PrisonerIepLevel) {
```
ushkarev, line 43: i get that we only exposed this for the special sync endpoint, but i do think there's a use case for deleting reviews.

what do you think about resurrecting this as a _new_ api like `DELETE /iep/reviews/booking/{bookingId}/id/{id}`? this method in particular has some business logic that's important to remember for the future (i.e. that there must be a "current" review)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveStoreService.kt
```diff
@@ -39,23 +38,4 @@ class IncentiveStoreService(
     log.debug("${savedReviews.count()} records saved")
     nextReviewDateUpdaterService.update(remainingBookingId)
   }
-
-  suspend fun patchIncentiveReview(
```
ushkarev, line 56: similar to the point above about allowing deletions, maybe patching is a useful public (non-sync) api to add

ushkarev (approved):
beautiful 🌟

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveStoreService.kt
ushkarev: don't mean it should be in this pr

xoen: I noticed these operation are now no longer possible. Deletion of an Incentive record always seemed iffy (at the very least I'd prefer a soft-delete), updating may be more useful (although when I checked these endpoints weren't used much).

However it's definitely out of scope. Plus even if adding an API it wouldn't still be possible at the moment. Which makes me think YAGNI.

PR #411, opened 2023-04-12
`/iep/` resources tidy up

xoen:
The old `IepLevelResource` class contained endpoints to manage Incentive records (`/iep/reviews/*`) but also an endpoint to get the list of Incentive levels for a given prison (`GET /iep/levels/{prisonId}`).

I've moved the latter in its own `IepLevelsResource` updating the Swagger name to reflect this and renamed the existing resource class to `IepReviewsResource`.

Relevant tests were also moved or removed to the correct test file.

**NOTE**: The API didn't change, the endpoints are working as before, this is an internal refactoring for clarity.
Also note that the `GET /iep/levels/{prisonId}` and corresponding resource will probably be deprecated and removed in favour of the new `/incentives/` endpoints once we migrated the reference data off NOMIS.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepReviewsResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepReviewsResource.kt
@@ -22,55 +22,15 @@ import uk.gov.justice.digital.hmpps.incentivesapi.dto.CurrentIepLevel
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.IepDetail
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.IepReview
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.IepSummary
-import uk.gov.justice.digital.hmpps.incentivesapi.dto.prisonapi.IepLevel
-import uk.gov.justice.digital.hmpps.incentivesapi.service.IepLevelService
 import uk.gov.justice.digital.hmpps.incentivesapi.service.PrisonerIepLevelReviewService
 import uk.gov.justice.digital.hmpps.incentivesapi.util.ensure
 
 @RestController
 @RequestMapping("/iep", produces = [MediaType.APPLICATION_JSON_VALUE])
 @Tag(name = "Incentive reviews", description = "Retrieve and add incentive review records. Ported from prison-api")
-class IepLevelResource(
-  private val iepLevelService: IepLevelService,
+class IepReviewsResource(
   private val prisonerIepLevelReviewService: PrisonerIepLevelReviewService,
 ) {
-  @GetMapping("/levels/{prisonId}")
-  @Operation(
-    summary = "Returns the valid IEP levels for specified prison",
-    description = "prison ID should be a 3 character string e.g. MDI = Moorland",
-    responses = [
-      ApiResponse(
-        responseCode = "200",
-        description = "IEP Level Information returned",
-      ),
-      ApiResponse(
-        responseCode = "400",
-        description = "Incorrect data specified to return IEP Level data",
-        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
-      ),
-      ApiResponse(
-        responseCode = "401",
-        description = "Unauthorized to access this endpoint",
-        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
-      ),
-      ApiResponse(
-        responseCode = "403",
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepLevelsResourceTest.kt
```diff
@@ -0,0 +1,70 @@
+package uk.gov.justice.digital.hmpps.incentivesapi.resource
+
+import kotlinx.coroutines.runBlocking
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import uk.gov.justice.digital.hmpps.incentivesapi.integration.IncentiveLevelResourceTestBase
+
+class IepLevelsResourceTest : IncentiveLevelResourceTestBase() {
```
xoen, line 9: FYI: I've copied the old `IepReviewsResourceTest.kt` file and removed all the tests and code related to the `/iep/reviews/*` endpoints.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepReviewsResource.kt
```diff
@@ -22,55 +22,15 @@ import uk.gov.justice.digital.hmpps.incentivesapi.dto.CurrentIepLevel
...
 
 @RestController
 @RequestMapping("/iep", produces = [MediaType.APPLICATION_JSON_VALUE])
 @Tag(name = "Incentive reviews", description = "Retrieve and add incentive review records. Ported from prison-api")
-class IepLevelResource(
-  private val iepLevelService: IepLevelService,
+class IepReviewsResource(
   private val prisonerIepLevelReviewService: PrisonerIepLevelReviewService,
 ) {
-  @GetMapping("/levels/{prisonId}")
```
xoen, line 37: FYI: Moved to `IepLevelsResource.kt` file.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepReviewsResource.kt
```diff
@@ -22,55 +22,15 @@ import uk.gov.justice.digital.hmpps.incentivesapi.dto.CurrentIepLevel
...
-import uk.gov.justice.digital.hmpps.incentivesapi.service.IepLevelService
 import uk.gov.justice.digital.hmpps.incentivesapi.service.PrisonerIepLevelReviewService
 import uk.gov.justice.digital.hmpps.incentivesapi.util.ensure
 
 @RestController
 @RequestMapping("/iep", produces = [MediaType.APPLICATION_JSON_VALUE])
 @Tag(name = "Incentive reviews", description = "Retrieve and add incentive review records. Ported from prison-api")
-class IepLevelResource(
-  private val iepLevelService: IepLevelService,
+class IepReviewsResource(
```
ushkarev, line 31: not too bothered, but maybe `IncentiveReviewsResource`? or something else that doesn't use the legacy term IEP?

ushkarev (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IepReviewsResource.kt
xoen: I deliberately kept the "iep" prefix for 2 reasons:
1. it's predictable and less confusing given the endpoints' URLs start with `/iep/reviews/` (and `/iep/levels`)
2. these endpoints are still using kinda legacy DTO/classes and we may want to rethink them (e.g. create new ones and deprecate these)

That's why I didn't want to create even more confusion by renaming this to something that may suggest these are "new" endpoints if that makes sense :)

ushkarev: 👌

xoen: (I'm also thinking these are really Incentive records given they may not necessarily be Incentive "reviews", but that's another story ☺️)

ushkarev: that's fair but IEP still boo ;)

xoen: Agreed, IEP boo :)

PR #414, opened 2023-04-13
INC-1197: Retain service methods to update/delete an Incentive record

xoen:
[We removed the `/iep/migration/*` and `/iep/sync/*` endpoints](https://github.com/ministryofjustice/hmpps-incentives-api/pull/409/) but Igor suggested to retain the functionality to update/delete an Incentive record in case we want to add new endpoints to do so. 

It's easier to revert and clean these commits now than in a few weeks or months.


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IncentiveRecordUpdate.kt
```diff
@@ -0,0 +1,18 @@
+package uk.gov.justice.digital.hmpps.incentivesapi.dto
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.v3.oas.annotations.media.Schema
+import java.time.LocalDateTime
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@Schema(description = "Request to update an Incentive record")
+data class IncentiveRecordUpdate(
```
xoen, line 9: It was `SyncPatchRequest`.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IncentiveRecordUpdate.kt
```diff
@@ -0,0 +1,18 @@
...
+  @Schema(description = "Updated date and time of the Incentive record", required = false, example = "2021-12-31T12:34:56.789012")
+  val reviewTime: LocalDateTime?,
```
xoen, line 11: It was `iepTime`, renamed for consistency with `PrisonerIepLevel`.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveStoreService.kt
```diff
@@ -38,4 +39,36 @@ class IncentiveStoreService(
...
+        prisonerIepLevelRepository.save(this.copy(current = true))
+      }
+    }
+  }
+
+  suspend fun updateIncentiveRecord(
+    update: IncentiveRecordUpdate,
+    incentiveRecord: PrisonerIepLevel,
+  ): PrisonerIepLevel {
+    update.current?.let {
```
ushkarev, line 60: i think we need this to be
```kotlin
if (update?.current) {
```
otherwise i'm pretty certain the block runs on any non-null value of current (including false)

xoen: I think you're right, and I think we had a similar problem before, I need to check tho.

ushkarev (approved):

PR #415, opened 2023-04-13
INC-1202: Restrict global incentive level deactivation

ushkarev:
To be allowed to deactivate a level globally, it must not be active in any prison

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepository.kt
@@ -97,4 +97,17 @@ interface PrisonIncentiveLevelRepository : CoroutineCrudRepository<PrisonIncenti
     """,
   )
   fun findPrisonIdsWithActiveLevels(): Flow<String>
+
+  /**
+   * All prisons that have this level active
+   */
+  @Query(
+    // language=postgresql
+    """
+    SELECT prison_id
+    FROM prison_incentive_level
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepository.kt
```diff
@@ -97,4 +97,17 @@ interface PrisonIncentiveLevelRepository : CoroutineCrudRepository<PrisonIncenti
...
+   */
+  @Query(
+    // language=postgresql
+    """
+    SELECT prison_id
+    FROM prison_incentive_level
+    WHERE active IS TRUE AND level_code = :levelCode
+    """,
+  )
+  fun findPrisonIdsWithActiveLevel(levelCode: String): Flow<String>
```
xoen, line 112: Could this just count the rows, we don't need the IDs don't we?

ushkarev: right now, you're right but i'm anticipating that i'll need the ids for future error message shown to users

xoen (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepository.kt
xoen: Don't mind too much in this case, I don't think it would be a big list.

PR #417, opened 2023-04-14
INC-1202: Restrict prison incentive level deactivation 

ushkarev:
To be allowed to deactivate a level in a prison, there must not be any prisoners currently on the level.
Offender search is used to get list of residents; this is matched against the incentives DB (which by definition is the source of truth on current levels).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/OffenderSearchService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/OffenderSearchService.kt
@@ -1,5 +1,7 @@
 package uk.gov.justice.digital.hmpps.incentivesapi.service
 
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
 import org.springframework.stereotype.Service
 import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.reactive.function.client.awaitBody
@@ -9,11 +11,10 @@ import uk.gov.justice.digital.hmpps.incentivesapi.dto.OffenderSearchPrisonerList
 @Service
 class OffenderSearchService(private val offenderSearchWebClient: WebClient) {
   /**
-   * Searches for offenders using a cell location prefix, e.g. MDI-1
-   * Requires role ROLE_PRISONER_IN_PRISON_SEARCH or ROLE_PRISONER_SEARCH
+   * Searches for offenders using a cell location prefix (e.g. MDI-1) returning a flow of pages
+   * Requires role PRISONER_IN_PRISON_SEARCH or PRISONER_SEARCH
    */
-  suspend fun findOffenders(prisonId: String, cellLocationPrefix: String): List<OffenderSearchPrisoner> {
-    val offenders = mutableListOf<OffenderSearchPrisoner>()
+  fun findOffendersAtLocation(prisonId: String, cellLocationPrefix: String): Flow<List<OffenderSearchPrisoner>> = flow {
     var page = 0
     do {
       val pageOfData = offenderSearchWebClient.get()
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/OffenderSearchService.kt
```diff
@@ -29,9 +30,20 @@ class OffenderSearchService(private val offenderSearchWebClient: WebClient) {
...
+
+  /**
+   * Searches for offenders using a cell location prefix (e.g. MDI-1) returning a complete list
+   * Requires role PRISONER_IN_PRISON_SEARCH or PRISONER_SEARCH
+   */
+  suspend fun getOffendersAtLocation(prisonId: String, cellLocationPrefix: String): List<OffenderSearchPrisoner> {
+    val offenders = mutableListOf<OffenderSearchPrisoner>()
+    findOffendersAtLocation(prisonId, cellLocationPrefix).collect {
+      offenders.addAll(it)
+    }
```
xoen, line 46: `.toList().flatten()` should do the same thing without having to use a mutable list (may be a bit more performant too?) - but check. There may be an even nicer way of doing this too (there are a bunch of `flat*` method on `Flow` but this seems clear enough)

ushkarev: >may be a bit more performant too?
i would expect the opposite! my version only creates 1 list and appends to it; to-list + flatten would first make a nested list and then do the same thing my version does

ushkarev: i looked around for a flatten-into-something type methods but didn't find any :(

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -376,10 +377,18 @@ class PrisonerIepLevelReviewService(
       log.info("No incentive records found for $removedPrisonerNumber, no records updated")
     }
   }
-}
 
-@OptIn(FlowPreview::class)
-fun <T> merge(vararg flows: Flow<T>): Flow<T> = flowOf(*flows).flattenMerge()
+  suspend fun countOfPrisonersOnLevelInPrison(prisonId: String, levelCode: String): Int {
+    return offenderSearchService.findOffendersAtLocation(prisonId, prisonId)
+      .fold(0) { countSoFar, offenders ->
```
xoen, line 383: 🤔 I wonder if you could change the method name to something like `anyPrisonerOnLevelInPrison()` (possibly a better name) and shortcut/return true this whenever you found a match...

this could save some HTTP requests/queries/iterations with a bit of luck as you only interested in knowing if there is any prisoner at that level to prevent the Prison Incentive level deactivation right?

ushkarev: i'm expecting to need the number returned for a future error message; if not then i agree with you! we can then use more performant SQL too

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/OffenderSearchService.kt
xoen: I guess my point was, I'd imagine library methods/potentially the compiler to be clever enough to "look" though these and be able to avoid unnecessary allocations/concatenations (and it reads better).

There was something like `flatMapMerge` but we don't need to transform it and I didn't quite got how it works tbh.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
ushkarev: @xoen actually, would you rather this method returned a boolean until we know that we really do want a number?

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/CountPrisonersServiceTest.kt
```diff
@@ -20,7 +20,7 @@ import java.time.ZoneId
 
 // TODO: move into inner class CountPrisonersOnLevel of PrisonerIepLevelReviewServiceTest once IepLevelService is removed
 class CountPrisonersServiceTest {
-  private val clock: Clock = Clock.fixed(Instant.ofEpochMilli(0), ZoneId.of("Europe/London"))
+  private val clock: Clock = Clock.fixed(Instant.parse("2022-08-01T12:45:00.00Z"), ZoneId.of("Europe/London"))
```
xoen, line 23: Shame, 70s were great from what I've heard 😝

xoen (dismissed):
Nice 👍

PR #419, opened 2023-04-15
Fix transactional annotation

ushkarev:
Spring only creates transactions when calling annotated methods on _other_ classes, invocations on `this` will not create transactions as the proxy object isn't used


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -344,7 +350,6 @@ class PrisonerIepLevelReviewService(
     )
   }
 
-  @Transactional
```
ushkarev, line 347: the only existing call to this method is from the `processOffenderEvent` in this class so the transactional annotation has no effect

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -135,6 +135,7 @@ class PrisonerIepLevelReviewService(
   suspend fun getReviewById(id: Long): IepDetail =
     prisonerIepLevelRepository.findById(id)?.toIepDetail(incentiveLevelService.getAllIncentiveLevelsMapByCode()) ?: throw NoDataFoundException(id)
 
+  @Transactional
```
ushkarev, line 138: adding annotation here so all effects of the received event are transactional

neilmendum (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
ushkarev: BUT if `createIepForReceivedPrisoner` or `mergedPrisonerDetails` were to raise domain events, then these might be received by another party prior to the transaction committing mean they get a stale read!

PR #422, opened 2023-04-18
Tweak Renovate config to only bump major versions of HMPPS Orb

neilmendum:
I attempted with https://github.com/ministryofjustice/hmpps-renovate-config/commit/00c1104a133de6d64207fb95af91bad410095a4a but doesn't seem to be working as per https://github.com/ministryofjustice/hmpps-incentives-api/pull/421

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@7.3.0
+  hmpps: ministryofjustice/hmpps@7
 
 parameters:
   alerts-slack-channel:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@7.3.0
+  hmpps: ministryofjustice/hmpps@7
```
xoen, line 4: 😬

xoen (approved):

.circleci/config.yml
ushkarev: 🤣

PR #424, opened 2023-04-18
Fix filtering offender search results when needing a whole prison's worth of people

ushkarev:
For a prison with fictional id AAI, the offender search endpoint works with `cellLocationPrefix` being `"AAI-"` or `""` but not `"AAI"`.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/CountPrisonersService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/CountPrisonersService.kt
@@ -15,7 +15,7 @@ class CountPrisonersService(
    * Returns the number of prisoners on given level at a prison
    */
   suspend fun countOfPrisonersOnLevelInPrison(prisonId: String, levelCode: String): Int {
-    return offenderSearchService.findOffendersAtLocation(prisonId, prisonId)
+    return offenderSearchService.findOffendersAtLocation(prisonId, "")
       .fold(0) { countSoFar, offenders ->
         val bookingIds = offenders.map { it.bookingId }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/CountPrisonersService.kt
```diff
@@ -15,7 +15,7 @@ class CountPrisonersService(
    * Returns the number of prisoners on given level at a prison
    */
   suspend fun countOfPrisonersOnLevelInPrison(prisonId: String, levelCode: String): Int {
-    return offenderSearchService.findOffendersAtLocation(prisonId, prisonId)
+    return offenderSearchService.findOffendersAtLocation(prisonId, "")
```
xoen, line 18: This is fine and I prefer it to pass `"{prisonId}-"` so don't feel like you have to change but I do wonder if have an optional parameter for the location (defaulting to `""`) would be slightly nicer on the consumer side of the service

xoen (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/CountPrisonersService.kt
ushkarev: yeah `"${prisonId}-"` felt more hacky to me

xoen: e.g. `offenderSearchService.findOffendersAtLocation("MDI")` conveys better what you're trying to do, e.g. get all prisoners in Moorland. But as I said, up to you.

xoen (approved):

PR #425, opened 2023-04-18
INC-1202: Check if some prisoner exists on a level in a prison

ushkarev:
Knowing the _number_ of prisoners on a level is not necessary so SQL and offender search calls can be made simpler.
Tweaks #417

PR #426, opened 2023-04-19
INC-1202: Return better error responses when incentive levels cannot be modified

ushkarev:
Error responses could always have contained an integer `errorCode` but it was not used in practice. For clients of the api to be able to _precisely_ know the error that occurred, it's better to use an enum / constant code rather than rely on text included in the message.

Similarly, the `moreInfo` field hasn't been used yet: Trying to remove an incentive level globally will now return an error with prisons that currently have it active.

Depends on (and won't merge with) #425

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/HmppsIncentivesApiExceptionHandler.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/HmppsIncentivesApiExceptionHandler.kt
@@ -71,6 +71,11 @@ class HmppsIncentivesApiExceptionHandler {
           },
           userMessage = "Validation failure: ${e.message}",
           developerMessage = e.message,
+          moreInfo = if (e is ValidationExceptionWithErrorCode) {
+            e.moreInfo
+          } else {
+            null
+          },
         ),
       )
   }
@@ -248,7 +253,7 @@ class ListOfDataNotFoundException(dataType: String, ids: Collection<Long>) :
 class DataIntegrityException(message: String) :
   Exception(message)
 
-class ValidationExceptionWithErrorCode(message: String, val errorCode: ErrorCode) :
+class ValidationExceptionWithErrorCode(message: String, val errorCode: ErrorCode, val moreInfo: String? = null) :
   ValidationException(message)
 
 enum class ErrorCode(val errorCode: Int) {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/HmppsIncentivesApiExceptionHandler.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/HmppsIncentivesApiExceptionHandler.kt
@@ -64,6 +64,11 @@ class HmppsIncentivesApiExceptionHandler {
       .body(
         ErrorResponse(
           status = BAD_REQUEST,
+          errorCode = if (e is ValidationExceptionWithErrorCode) {
+            e.errorCode
+          } else {
+            null
+          },
           userMessage = "Validation failure: ${e.message}",
           developerMessage = e.message,
         ),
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/HmppsIncentivesApiExceptionHandler.kt
```diff
@@ -261,14 +260,13 @@ class ListOfDataNotFoundException(dataType: String, ids: Collection<Long>) :
...
 
-@JsonInclude(JsonInclude.Include.NON_NULL)
-@Schema(description = "Error Response")
+@Schema(description = "Error response")
 data class ErrorResponse(
-  @Schema(description = "Status of Error", example = "500", required = true)
+  @Schema(description = "HTTP status code", example = "500", required = true)
   val status: Int,
-  @Schema(description = "Error Code", example = "500", required = false)
+  @Schema(description = "Internal error code", example = "123", required = false)
```
xoen, line 293: I can't think of a better description atm but is this internal if the code is used by the UI/possibly other clients 🤔? 

Maybe something like "The error code could be used to discriminate between different errors, e.g. for validation purposes" 🤷

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/helper/ErrorResponseHelper.kt
```diff
@@ -4,7 +4,7 @@ import org.assertj.core.api.Assertions.assertThat
 import org.springframework.http.HttpStatus
 import org.springframework.test.web.reactive.server.WebTestClient
 
-fun <T : WebTestClient.ResponseSpec> T.expectErrorResponse(status: HttpStatus, vararg messages: String): T {
+fun <T : WebTestClient.ResponseSpec> T.expectErrorResponse(status: HttpStatus, vararg messages: String, errorCode: Int? = null): T {
```
xoen: could `errorCode` have `ErrorCode?` type? (but still check the value/number?)

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveLevelResourceTest.kt
```diff
@@ -320,7 +326,11 @@ class IncentiveLevelResourceTest : IncentiveLevelResourceTestBase() {
           """,
         )
         .exchange()
-        .expectErrorResponse(HttpStatus.BAD_REQUEST, "Incentive level with code STD already exists")
+        .expectErrorResponse(
+          HttpStatus.BAD_REQUEST,
+          "Incentive level with code STD already exists",
+          errorCode = 102,
```
xoen, line 332: I think pass the enum value would be much more readable :)

xoen (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/helper/ErrorResponseHelper.kt
ushkarev: this was actually deliberate cos those numbers in the enum are too easy to change. if someone in future adds a new constant to the error codes and decides to renumber things, they'll get a lot of tests failing. i wanted it to be clear that you break something if you change the numbers.

ushkarev: i agree that it's not as elegant and makes the tests more tedious to read and write, but i think they're a bit safer that way… what do you think now?

xoen: 🤔 I suppose there is always a risk of something changing but with these enums they should be considered as part of the API contract almost.

Maybe there should be tests to check the code and the type/value match and then use the enum everywhere else 🤷 but maybe it feels like faff?

xoen (approved):

PR #432, opened 2023-04-22
INC-1234: Send audit events for other levels changed when one becomes the new default

ushkarev:
Previously, only the level that was being directly edited had events published for the change. Now, if another level becomes non-default-for-admissions, it will also publish audit events.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepository.kt
@@ -1,7 +1,6 @@
 package uk.gov.justice.digital.hmpps.incentivesapi.jpa.repository
 
 import kotlinx.coroutines.flow.Flow
-import org.springframework.data.r2dbc.repository.Modifying
 import org.springframework.data.r2dbc.repository.Query
 import org.springframework.data.repository.kotlin.CoroutineCrudRepository
 import org.springframework.stereotype.Repository
@@ -74,16 +73,16 @@ interface PrisonIncentiveLevelRepository : CoroutineCrudRepository<PrisonIncenti
    * Sets levels other than the one privided to be non-default for this prison
    * Used to ensure there is only one default level for admission in a prison
    */
-  @Modifying
   @Query(
     // language=postgresql
     """
     UPDATE prison_incentive_level
     SET default_on_admission = FALSE
-    WHERE prison_id = :prisonId AND NOT level_code = :levelCode
+    WHERE prison_id = :prisonId AND NOT level_code = :levelCode AND default_on_admission IS TRUE
+    RETURNING level_code
     """,
   )
-  suspend fun setOtherLevelsNotDefaultForAdmission(prisonId: String, levelCode: String): Int?
+  fun setOtherLevelsNotDefaultForAdmission(prisonId: String, levelCode: String): Flow<String>
 
   /**
    * All prisons that have active level configurations, effectively a way to find all active prisons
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepository.kt
```diff
@@ -74,16 +73,16 @@ interface PrisonIncentiveLevelRepository : CoroutineCrudRepository<PrisonIncenti
    * Sets levels other than the one privided to be non-default for this prison
    * Used to ensure there is only one default level for admission in a prison
    */
-  @Modifying
```
xoen, line 77: why no longer `@Modifying`?

ushkarev: it's a slightly strange annotation: it tells spring to only care about the number or rows changed so the `… RETURNING ???` is ignored if you use the annotation.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepository.kt
```diff
@@ -74,16 +73,16 @@ interface PrisonIncentiveLevelRepository : CoroutineCrudRepository<PrisonIncenti
...
     // language=postgresql
     """
     UPDATE prison_incentive_level
     SET default_on_admission = FALSE
-    WHERE prison_id = :prisonId AND NOT level_code = :levelCode
+    WHERE prison_id = :prisonId AND NOT level_code = :levelCode AND default_on_admission IS TRUE
+    RETURNING level_code
     """,
   )
-  suspend fun setOtherLevelsNotDefaultForAdmission(prisonId: String, levelCode: String): Int?
```
xoen, line 86: Is it worth document this returns the updated level codes (one really)?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonIncentiveLevelRepository.kt
xoen: Weird, oh I see it means "this is an update query, ignore the returned value" 👌

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/PrisonIncentiveLevelResourceTest.kt
```diff
@@ -306,21 +308,35 @@ class PrisonIncentiveLevelResourceTest : IncentiveLevelResourceTestBase() {
...
+
+      val expectedPrisonIncentiveLevelChanges = 2
+      assertThat(domainEvents).hasSize(expectedPrisonIncentiveLevelChanges)
+      assertThat(auditMessages).hasSize(expectedPrisonIncentiveLevelChanges)
+      assertThat(domainEvents.count { it.eventType == "incentives.prison-level.changed" }).isEqualTo(expectedPrisonIncentiveLevelChanges)
+      assertThat(auditMessages.count { it.what == "PRISON_INCENTIVE_LEVEL_UPDATED" }).isEqualTo(expectedPrisonIncentiveLevelChanges)
+
+      assertThat(domainEvents).allMatch {
+        it.additionalInformation?.prisonId == "WRI" &&
+          it.description.matches(Regex("^Incentive level \\(...\\) in prison WRI has been updated$"))
```
xoen, line 322: nice

xoen (approved):

PR #440, opened 2023-05-11
Use newer (and slimmer) version of localstack for testing and running locally

ushkarev:
…removing irrelevant and deprecated configuration


ushkarev:
replaced by #545

PR #447, opened 2023-05-23
INC-1147 Remove feature flag for retrieving reference data from NOMIS (via prison API)

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/config/ApplicationInsightsConfiguration.kt
```diff
@@ -12,5 +12,3 @@ class ApplicationInsightsConfiguration {
   @Bean
   fun telemetryClient(): TelemetryClient = TelemetryClient()
 }
-
-fun TelemetryClient.trackEvent(name: String, properties: Map<String, String>) = this.trackEvent(name, properties, null)
```
ushkarev, line 16: feels like we actually still need this?
https://github.com/ministryofjustice/hmpps-incentives-api/blob/427c845be71c1e54d9a86d1a9f35627a482e0942/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/AuditService.kt#L49-L53

ushkarev: oh no, wait, that uses the 3rd param so ignore me

ushkarev (approved):

PR #449, opened 2023-05-23
INC-1251: Refactor legacy resource endpoint and services

ushkarev:
• `/incentive/prison-levels/{prisonId}` should be used in place of `/iep/levels/{prisonId}` (which initially had been moved from prison-api and is now outdated) and improve documentation for the deprecated endpoint
• Replace calls to `IepLevelService.getIepLevelsForPrison` with `PrisonIncentiveLevelService.getActivePrisonIncentiveLevels` as the former is just a façade for the latter
• Rename `IepLevelService` to `NearestPrisonIncentiveLevelService` so that it’s purpose is clearer
• Fix version of `@stoplight/spectral-cli` as the latest is currently broken


.circleci/config.yml
```diff
@@ -86,15 +86,17 @@ jobs:
       - checkout
       - attach_workspace:
           at: ~/docs
-      # Create a folder for results to live in
-      - run: "[ -d lint-results ] || mkdir lint-results"
+      - run:
+          name: Create a folder for results to live in
+          command: "[ -d build/lint-results ] || mkdir -p build/lint-results"
```
ushkarev, line 91: might as well place under `./build` which is git-ignored and where other test results are saved

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveStoreService.kt
```diff
@@ -57,7 +57,7 @@ class IncentiveStoreService(
     update: IncentiveRecordUpdate,
     incentiveRecord: PrisonerIepLevel,
   ): PrisonerIepLevel {
-    if (update?.current == true) {
+    if (update.current == true) {
```
ushkarev, line 60: `update` is not an optional type

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
```diff
@@ -40,9 +29,11 @@ class IepLevelService(
    * person’s level when they are transferred, we need to find the "nearest highest" level
    * according to HMPPS policy. Falls back to prison's default level when there's nothing else to do.
    */
-  suspend fun findNearestHighestLevel(prisonId: String, prisonLevels: List<IepLevel>, levelCode: String): String {
+  suspend fun findNearestHighestLevel(prisonId: String, prisonLevels: List<PrisonIncentiveLevel>, levelCode: String): String {
     val defaultLevelCode = chooseDefaultLevel(prisonId, prisonLevels)
-    val levelCodesAvailableInPrison = prisonLevels.filter(IepLevel::active).map(IepLevel::iepLevel).toSet()
+    val levelCodesAvailableInPrison = prisonLevels.filter(PrisonIncentiveLevel::active)
+      .map(PrisonIncentiveLevel::levelCode)
+      .toSet()
```
xoen, line 36: Similar comment to above, if this service used the `PrisonIncentiveLevelService` it could easily get the available active levels and the default one (unless add that service to this would result in a cycle I'm not aware of, I didn't try)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -213,8 +214,8 @@ class PrisonerIepLevelReviewService(
   }
 
   private suspend fun getIepLevelForReviewType(prisonerInfo: PrisonerAtLocation, reviewType: ReviewType): String {
-    val iepLevelsForPrison = iepLevelService.getIepLevelsForPrison(prisonerInfo.agencyId)
-    val defaultLevelCode = iepLevelService.chooseDefaultLevel(prisonerInfo.agencyId, iepLevelsForPrison)
+    val prisonIncentiveLevels = prisonIncentiveLevelService.getActivePrisonIncentiveLevels(prisonerInfo.agencyId)
+    val defaultLevelCode = iepLevelService.chooseDefaultLevel(prisonerInfo.agencyId, prisonIncentiveLevels)
```
xoen: We could stop using `chooseDefaultLevel()`. Wouldn't `prisonIncentiveLevels .first(PrisonIncentiveLevel::defaultOnAdmission).levelCode` be equivalent now that prisons must have a default level?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIepLevelReviewService.kt
```diff
@@ -230,7 +231,7 @@ class PrisonerIepLevelReviewService(
           val levelCodeBeforeTransfer =
             iepHistory.sortedBy(IepDetail::iepTime).lastOrNull { it.agencyId != prisonerInfo.agencyId }?.iepCode
               ?: defaultLevelCode // if no previous prison
-          iepLevelService.findNearestHighestLevel(prisonerInfo.agencyId, iepLevelsForPrison, levelCodeBeforeTransfer)
+          iepLevelService.findNearestHighestLevel(prisonerInfo.agencyId, prisonIncentiveLevels, levelCodeBeforeTransfer)
```
xoen: Similarly, the list of active prison levels could happen inside that service/method.

It would simplify the method invocation, but maybe it's annoying to change?

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelServiceTest.kt
```diff
@@ -210,9 +124,9 @@ class IepLevelServiceTest {
     fun `falls back to the prison's default when target is not found`(): Unit = runBlocking {
```
xoen: Is it still possible to not have a default level in a prison? 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/LegacyPrisonIncentiveLevelResource.kt
```diff
@@ -11,31 +11,25 @@ import org.springframework.web.bind.annotation.PathVariable
...
 import uk.gov.justice.digital.hmpps.incentivesapi.service.PrisonIncentiveLevelAuditedService
-import uk.gov.justice.digital.hmpps.incentivesapi.util.ensure
 
 @RestController
 @RequestMapping("/iep", produces = [MediaType.APPLICATION_JSON_VALUE])
-@Tag(name = "Prison incentive levels", description = "Retrieve incentive levels for a prison")
+@Tag(name = "[Legacy] Prison incentive levels", description = "Incentive levels available in prisons")
 @Deprecated("Use `/incentive/prison-levels/…`")
-class IepLevelsResource(
+class LegacyPrisonIncentiveLevelResource(
```
xoen, line 21: I like the ugly name to really make the point this shouldn't be used, even if this is just an hint for us internally ❤️

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
```diff
@@ -2,35 +2,24 @@ package uk.gov.justice.digital.hmpps.incentivesapi.service
...
-  fun chooseDefaultLevel(prisonId: String, prisonLevels: List<IepLevel>): String {
-    val activePrisonLevels = prisonLevels.filter(IepLevel::active)
-    return activePrisonLevels.firstOrNull(IepLevel::default)?.iepLevel
+  /**
+   * Selects the default level (for new admissions) for a prison
+   * NB: this existed largely for compatibility with NOMIS which allowed there to be no default set,
+   * business rules now prevent this from happening
+   */
+  fun chooseDefaultLevel(prisonId: String, prisonLevels: List<PrisonIncentiveLevel>): String {
+    val activePrisonLevels = prisonLevels.filter(PrisonIncentiveLevel::active)
```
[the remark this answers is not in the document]
ushkarev: now, the whole method could probably be removed as there's _always_ a default. i was being a bit conservative with this PR because the data wasn't fixed in production until after this morning's go-live

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
ushkarev: inactive levels aren't even passed into this method anymore. this entire class could be simplified a lot but i was wondering if that ought to be left for a slightly later pr once we're convinced that the data is all good in prod…?

now:
1) there will always be 1 default
2) there will not be inactive levels passed into either method in this class

previously, neither were guaranteed

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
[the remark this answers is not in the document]
xoen: Yeah that's what I'm thinking about the first part 🤔 

I think what you did is fine, but if you fancy I think `chooseDefaultLevel()` could be possibly removed/made private or at least simplified.

Similarly the class could use `PrisonIncentiveLevelService` directly and don't have to rely on client code to pass the list of levels (the client code currently use this service to get the list of prison levels).

xoen (approved):
Nice ❤️ 

I left some comments as potentially we could simplify `NearestPrisonIncentiveLevelService` to use `PrisonIncentiveLevelService` directly instead of relying on the client code to pass in the list of levels for an establishment, etc.

But don't have to be done in this PR or even at all.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IepLevelService.kt
ushkarev: should hopefully be addressed by https://github.com/ministryofjustice/hmpps-incentives-api/pull/452

PR #452, opened 2023-05-24
INC-1251: Simplify code now that there is a guaranteed default level in every prison

ushkarev:
Business rules guarantee that each prison has exactly 1 default level for new admissions so the api will no longer try to fall back to Standard or first-available level; it'll be considered a data integrity error (with a 500 response).

Addresses comments in #449


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveLevelService.kt
```diff
@@ -26,7 +26,7 @@ import uk.gov.justice.digital.hmpps.incentivesapi.dto.IncentiveLevelUpdate as In
 class IncentiveLevelService(
   private val clock: Clock,
   private val incentiveLevelRepository: IncentiveLevelRepository,
-  private val prisonIncentiveLevelService: PrisonIncentiveLevelService,
+  private val prisonIncentiveLevelService: PrisonIncentiveLevelService, // OK to inject unaudited version as audited subclass passes in audited form instead
```
xoen, line 29: Seems like rule of thumb is "always use the audited version" - could this service also use the `PrisonIncentiveLevelAuditedService` one too?

ushkarev: the subclass does this:
https://github.com/ministryofjustice/hmpps-incentives-api/blob/25147e7aa62c25cedcabfe6c2a4130fa9a4839db/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveLevelAuditedService.kt#L25

_this_ class (`IncentiveLevelService`) is unaudited so does not need to require _only_ the audited interface of a peer dependency.

xoen (approved):
Nice to see some old code go away and the remaining one being simplified :)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveLevelService.kt
xoen: I suppose I was thinking of the two services as separate (incentive level vs prison incentive level one) but I guess makes sense to use the unaudited prison level service in this context of unaudited incentive level service.

xoen: Also, good point about the "does not need to require _only_ the audited" one - I forgot you wired things in such a way you could use the audited version when a simple version is required (in other words an instance of `PrisonIncentiveLevelService` is-a `PrisonIncentiveLevelService`)

PR #453, opened 2023-05-24
INC-1265: Add api endpoint that deactivates all incentive levels in a prison

ushkarev:
This is needed when a prison is closed and to tidy up ancient prison incentive level information from long-closed prisons.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonIncentiveLevelAuditedService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonIncentiveLevelAuditedService.kt
@@ -38,6 +38,13 @@ class PrisonIncentiveLevelAuditedService(
       }
   }
 
+  override suspend fun deactivateAllPrisonIncentiveLevels(prisonId: String): List<PrisonIncentiveLevelDTO> {
+    return super.deactivateAllPrisonIncentiveLevels(prisonId)
+      .onEach {
+        auditAndPublishPrisonLevelChangeEvent(it)
+      }
+  }
+
   private suspend fun auditAndPublishPrisonLevelChangeEvent(prisonIncentiveLevel: PrisonIncentiveLevelDTO) {
     auditService.sendMessage(
       AuditType.PRISON_INCENTIVE_LEVEL_UPDATED,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonIncentiveLevelService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonIncentiveLevelService.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.incentivesapi.service
 
 import jakarta.validation.ValidationException
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.toList
 import org.springframework.stereotype.Service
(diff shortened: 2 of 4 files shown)


Mjwillis (approved):
Lgtm

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/PrisonIncentiveLevelResourceTest.kt
```diff
@@ -1216,6 +1216,297 @@ class PrisonIncentiveLevelResourceTest : IncentiveLevelResourceTestBase() {
...
+      runBlocking {
+        assertThat(prisonIncentiveLevelRepository.count()).isZero
+      }
+
+      assertNoDomainEventSent()
+      assertNoAuditMessageSent()
+    }
+
+    @Test
+    fun `deactivates all prison incentive levels when all are already inactive`() {
```
xoen, line 1244: This test case made me wonder if the endpoint should return all prison incentive levels or only the updated/deactivated ones.

Especially as the `incentives.prison-level.changed` domain event is published even for levels which were already inactive and didn't change (if I understand the current logic correctly).

Don't know if it would be better to only return the prison levels which were active/saved instead? Don't know if you thought about this already.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonIncentiveLevelAuditedService.kt
```diff
@@ -38,6 +38,13 @@ class PrisonIncentiveLevelAuditedService(
       }
   }
 
+  override suspend fun deactivateAllPrisonIncentiveLevels(prisonId: String): List<PrisonIncentiveLevelDTO> {
+    return super.deactivateAllPrisonIncentiveLevels(prisonId)
+      .onEach {
+        auditAndPublishPrisonLevelChangeEvent(it)
+      }
```
xoen, line 45: As the `deactivateAllPrisonIncentiveLevels()` currently returns all levels for a prison - even the one which were already inactive - this would publish 'changed' domain events for prison levels that didn't really change.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonIncentiveLevelService.kt
```diff
@@ -167,6 +168,31 @@ class PrisonIncentiveLevelService(
...
+        moreInfo = prisonIncentiveLevelsWithPrisoners.map(PrisonIncentiveLevel::levelCode).joinToString(","),
+      )
+    }
+    prisonIncentiveLevelRepository.saveAll(
+      prisonIncentiveLevels.filter(PrisonIncentiveLevel::active)
+        .map { prisonIncentiveLevel ->
+          prisonIncentiveLevel.withUpdate(PrisonIncentiveLevelUpdateDTO(active = false))
+        },
+    ).collect()
+    return getAllPrisonIncentiveLevels(prisonId)
```
xoen, line 193: Maybe if this returns the result of the `collect()` instead of re-reading the list of prison levels, it means `incentives.prison-level.changed` domain events would be published only for prison levels that were actually inactivated.

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/IncentiveLevelResourceTest.kt
```diff
@@ -28,17 +28,13 @@ class IncentiveLevelResourceTest : IncentiveLevelResourceTestBase() {
   }
 
   @Nested
-  inner class `retrieving incentive levels` {
```
xoen, line 31: Would be better to group tests by the endpoint tested rather than RO/RW? Too granular?

xoen: Ignore me, watching the whole diff with whitespaces just confused me 😅

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonIncentiveLevelAuditedService.kt
ushkarev: i figured that a few superfluous change events would not a big deal. given that the api’s caller wants the whole set to be inactive, it’s not bad to pretend that they all changed even if some did not. anyway, it’s addressed now :)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonIncentiveLevelService.kt
ushkarev: ok, changed in eb64ca910eed378464c11acc390ae5953f1124be – now the service only returns the changed levels. the api endpoint still returns the full set given the caller wanted the _whole_ prison deactivated.

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/PrisonIncentiveLevelResourceTest.kt
xoen: **Update**: Discussed and decided that:
- domain events will be published only for prison levels that actually changed
- endpoint will return the complete list of levels for an enstiblishment

PR #456, opened 2023-05-30
INC-1269: Increase test coverage for the case of a prison being re-activated

ushkarev:
When a prison is closed, it's prison incentive levels should be deactivated. This tests the situation where the same prison is later reopened.
NB: the first level that is activated must be a default according to business rules, but furthermore it must also have been the level that was _previously_ the default.

If a prison had these levels:
• BAS active
• STD active
• ENH active and default
and is closed, then it ends up with:
• BAS inactive
• STD inactive
• ENH inactive and default

To reactivate the whole prison the actions needed are:

1) activate ENH resulting in 
• BAS inactive
• STD inactive
• ENH active and default

2) activate all other required levels (BAS & STD) resulting in 
• BAS active
• STD active
• ENH active and default

only _then_ can the default level be changed if necessary.

Changing the default level for a prison when the current default is inactive is not always possible. Changing the default level necessarily requires a side effect of making the previous level no longer default. If that previous level was globally required, it will restrict updating it as inactive.

Relates to [ui#519](https://github.com/ministryofjustice/hmpps-incentives-ui/pull/519) which will now try activating the previously default level first.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/PrisonIncentiveLevelResourceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/PrisonIncentiveLevelResourceTest.kt
@@ -906,6 +906,84 @@ class PrisonIncentiveLevelResourceTest : IncentiveLevelResourceTestBase() {
       }
 
       @Test
+      fun `partially updates prison incentive levels if all being activated one-by-one`() {
+        makePrisonIncentiveLevel("MDI", "BAS")
+        makePrisonIncentiveLevel("MDI", "STD") // Standard was the default for admission before deactivation
+        makePrisonIncentiveLevel("MDI", "ENH")
+        runBlocking {
+          prisonIncentiveLevelRepository.saveAll(
+            prisonIncentiveLevelRepository.findAllByPrisonId("MDI").map {
+              it.copy(active = false)
+            }.toList(),
+          ).collect()
+        }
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/PrisonIncentiveLevelResourceTest.kt
```diff
@@ -905,6 +905,45 @@ class PrisonIncentiveLevelResourceTest : IncentiveLevelResourceTestBase() {
...
+        makePrisonIncentiveLevel("MDI", "ENH")
+        runBlocking {
+          prisonIncentiveLevelRepository.saveAll(
+            prisonIncentiveLevelRepository.findAllByPrisonId("MDI").map {
+              it.copy(active = false)
+            }.toList(),
+          ).collect()
+        }
+
+        // NB: the previously-default level must be acticated first to conform to business rules
```
xoen, line 921: activated

ushkarev: "acticated" should be a word :p

xoen: it should!

xoen (approved):

PR #461, opened 2023-06-16
INC-1276: Add api endpoint to reset levels in a prison

ushkarev:
This is useful when setting up a new prison. If in future there are domain events raised when a prison is created, incentives-api could listen for them and automatically set up the prison.

It can also be used to ensure that a prison has the required set of levels active and that Standard is the default for admission.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonIncentiveLevelService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonIncentiveLevelService.kt
@@ -228,7 +228,7 @@ class PrisonIncentiveLevelService(
     prisonId: String,
     levelCode: String,
   ): PrisonIncentiveLevel {
-    val fallbackSpendLimits = spendLimitPolicy.getOrDefault(levelCode, spendLimitPolicy["ENH"]!!)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonIncentiveLevelService.kt
```diff
@@ -257,15 +298,18 @@ class PrisonIncentiveLevelService(
...
 )
 
-private val spendLimitPolicy = mapOf(
-  "BAS" to SpendLimits(27_50, 275_00, 5_50, 55_00),
-  "STD" to SpendLimits(60_50, 605_00, 19_80, 198_00),
-  "ENH" to SpendLimits(66_00, 660_00, 33_00, 330_00),
+private val prisonIncentiveLevelPolicies = mapOf(
+  "BAS" to Policy(27_50, 275_00, 5_50, 55_00, 2, 1),
+  "STD" to Policy(60_50, 605_00, 19_80, 198_00, 2, 1),
+  "ENH" to Policy(66_00, 660_00, 33_00, 330_00, 2, 1),
```
ushkarev, line 314: Note that the policy document is not explicit about how many visit orders each level should allow. This PR does not change the "defaults" that were already set.

Mjwillis (approved):
LGTM

PR #466, opened 2023-07-28
Fix location lookup when retrieving reviews

ushkarev:
We want to be able to look up reviews for a given location by using a prefix like "MDI-1-" (including a trailing hyphen) so that other locations aren't accidentally included. Using a prefix of "MDI-1" (no trailing hyphen) would also match locations starting with "MDI-10" and that's wrong.

However, when looking up the details of a location in prison-api, trailing hyphens will not be valid so need to be removed.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsServiceTest.kt
@@ -142,7 +142,7 @@ class IncentiveReviewsServiceTest {
         prisonId = "MDI",
       ),
     )
-    whenever(prisonApiService.getLocation(any())).thenReturnLocation("MDI-2-1")
+    whenever(prisonApiService.getLocation(any())).thenReturnLocation("MDI-2")
     whenever(offenderSearchService.getOffendersAtLocation(any(), any())).thenReturn(offenders)
     val nextReviewDatesMap = mapOf(
       110001L to LocalDate.now(clock).plusYears(1),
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/util/ErrorCodeTest.kt
```diff
@@ -7,7 +7,7 @@ import uk.gov.justice.digital.hmpps.incentivesapi.config.ErrorCode
 class ErrorCodeTest {
   @Test
   fun `error codes should all be unique`() {
-    val errorCodes = ErrorCode.values()
+    val errorCodes = ErrorCode.entries
```
ushkarev, line 10: cf. https://kotlinlang.org/docs/whatsnew19.html#stable-replacement-of-the-enum-class-values-function

ashleygyngell (approved):

ashleygyngell (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsServiceTest.kt
```diff
@@ -142,17 +142,18 @@ class IncentiveReviewsServiceTest {
...
       110001L to LocalDate.now(clock).plusYears(1),
       110002L to LocalDate.now(clock).plusYears(1),
     )
     whenever(nextReviewDateGetterService.getMany(offenders)).thenReturn(nextReviewDatesMap)
 
-    val reviews = incentiveReviewsService.reviews("MDI", "MDI-2-1", "STD")
+    val reviews = incentiveReviewsService.reviews("MDI", "MDI-2-", "STD")
 
-    verify(offenderSearchService, times(1)).getOffendersAtLocation(any(), eq("MDI-2-1"))
+    verify(offenderSearchService, times(1)).getOffendersAtLocation(any(), eq("MDI-2-"))
```
ushkarev, line 155: we want prisoners whose cell locations starts with the actual query provided, but…

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/IncentiveReviewsServiceTest.kt
```diff
@@ -142,17 +142,18 @@ class IncentiveReviewsServiceTest {
...
+    verify(prisonApiService, times(1)).getLocation(eq("MDI-2"))
```
ushkarev, line 156: …we want to look up the location trimming the trailing hyphen (else it's not a real location ID)

PR #503, opened 2023-10-26
INC-1321: Automate KPIs

xoen:
Monthly task to calculate/save:
1. the number of reviews occurred in the previous, complete, month
2. the number of prisoners reviewed in the previous, complete, month
3. the number of prisoners overdue a review at the point when the task runs

First 2 can be calculate with a single query that look at the reviews in the previous month.
3rd is a bit more complicated because it gets the data from the DB but then it filter out people no longer in prison using Prison API/Offender Search API.

The automated task runs on the 1st of the month at 3:00am in production.
It also run on the 1st working day of the month at 10:00am in preprod/prod (because Offender Search API is switched off in dev/preprod during non-working hours).

Shedlock is used to prevent the task from running multiple times in all the application instances.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/HmppsIncentivesApi.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/HmppsIncentivesApi.kt
@@ -1,7 +1,10 @@
 package uk.gov.justice.digital.hmpps.incentivesapi
 
+import net.javacrumbs.shedlock.spring.annotation.EnableSchedulerLock
 import org.springframework.boot.autoconfigure.SpringBootApplication
 import org.springframework.boot.runApplication
+import org.springframework.scheduling.annotation.EnableScheduling
+import uk.gov.justice.digital.hmpps.incentivesapi.config.defaultLockAtLeastFor
 
 /**
  * Used as the system oauth username when calling HMPPS apis
@@ -10,6 +13,8 @@ import org.springframework.boot.runApplication
 const val SYSTEM_USERNAME = "INCENTIVES_API"
 
 @SpringBootApplication
+@EnableScheduling
+@EnableSchedulerLock(defaultLockAtMostFor = defaultLockAtLeastFor)
 class HmppsIncentivesApi
 
 fun main(args: Array<String>) {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/PrisonAPI.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/PrisonAPI.kt
@@ -8,7 +8,7 @@ import java.time.LocalDateTime
 data class Prison(
   val agencyId: String,
   val description: String,
-  val longDescription: String,
+  val longDescription: String?,
   val agencyType: String,
   val active: Boolean,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/PrisonAPI.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/PrisonAPI.kt
@@ -5,6 +5,14 @@ import uk.gov.justice.digital.hmpps.incentivesapi.service.PrisonerInfoForNextRev
 import java.time.LocalDate
 import java.time.LocalDateTime
 
+data class Prison(
+  val agencyId: String,
+  val description: String,
+  val longDescription: String,
+  val agencyType: String,
+  val active: Boolean,
+)
+
 data class PrisonerAtLocation(
   val bookingId: Long,
   val facialImageId: Long,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/Kpi.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/Kpi.kt
@@ -1,0 +1,30 @@
+package uk.gov.justice.digital.hmpps.incentivesapi.jpa
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.data.annotation.Id
+import org.springframework.data.annotation.ReadOnlyProperty
+import org.springframework.data.annotation.Transient
+import org.springframework.data.domain.Persistable
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+data class Kpi(
+  @Id
+  val day: LocalDate,
+
+  val overdueReviews: Int,
+  val previousMonthReviewsConducted: Int,
+  val previousMonthPrisonersReviewed: Int,
+
+  @ReadOnlyProperty
+  val whenCreated: LocalDateTime? = null,
+
+  val whenUpdated: LocalDateTime? = null,
+
+  @Transient
+  @Value("false")
+  val new: Boolean = true,
+) : Persistable<LocalDate> {
+  override fun getId(): LocalDate = day
+  override fun isNew(): Boolean = new
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/KpiRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/KpiRepository.kt
@@ -12,10 +12,6 @@ data class ReviewsConductedPrisonersReviewed(
   val prisonersReviewed: Int,
 )
 
-data class PrisonerNumberOverdue(
-  val prisonerNumber: String,
-)
-
 @Repository
 interface KpiRepository : CoroutineCrudRepository<Kpi, LocalDate> {
   /**
@@ -58,5 +54,5 @@ interface KpiRepository : CoroutineCrudRepository<Kpi, LocalDate> {
     ;
     """,
   )
-  fun getPrisonerNumbersOverdueReview(): Flow<PrisonerNumberOverdue>
+  fun getPrisonerNumbersOverdueReview(): Flow<String>
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/KpiRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/KpiRepository.kt
@@ -46,10 +46,11 @@ interface KpiRepository : CoroutineCrudRepository<Kpi, LocalDate> {
     //language=postgresql
     """
     -- this may include people no longer in prison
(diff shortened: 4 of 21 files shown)


src/main/resources/application.yml
```diff
@@ -2,6 +2,11 @@ info.app:
   name: Hmpps Incentives API
   version: 1.0
 
+task:
+  update-kpis:
+    # Every 1st of the month at 3:00am
```
ushkarev, line 7: oh actually sounds v sensible to do the "how many overdue reviews?" check at night cos nobody is part-way through whatever they've planned for the day.

xoen: Yeah, and it's a snapshot anyway so as long as it's a number taken around the same time it doesn't really matter the precise time when it's taken.

BTW: I had a chat with Steve this morning and he concurs. He's happy with this and he agrees it doesn't have to be taken separately on the last day of the month.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/PrisonAPI.kt
```diff
@@ -5,6 +5,14 @@ import uk.gov.justice.digital.hmpps.incentivesapi.service.PrisonerInfoForNextRev
 import java.time.LocalDate
 import java.time.LocalDateTime
 
+data class Prison(
+  val agencyId: String,
+  val description: String,
+  val longDescription: String,
```
ushkarev, line 11: potential danger: this field in nullable (at least according to swagger docs; i didn't check source)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonApiService.kt
```diff
@@ -73,4 +74,12 @@ class PrisonApiService(
       .retrieve()
       .awaitBody()
   }
+
+  suspend fun getActivePrisons(useClientCredentials: Boolean = false): List<Prison> {
```
ushkarev, line 78: `useClientCredentials` is a bit of a mystery to me, but i know there's an effort to use "system" tokens and their roles. just leaving this note here to maybe explore using the `hmpps-incentives-api` token and giving it any extra roles if necessary

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/Kpi.kt
```diff
@@ -0,0 +1,30 @@
...
+data class Kpi(
+  @Id
+  val day: LocalDate,
+
+  val overdueReviews: Int,
+  val previousMonthReviewsConducted: Int,
+  val previousMonthPrisonersReviewed: Int,
+
+  @ReadOnlyProperty
+  val whenCreated: LocalDateTime? = null,
```
ushkarev, line 20: from recollection, our "reactive" setup means you need to ` = LocalDateTime.now()` otherwise whenCreated and whenUpdated are not set…? or did it work for you?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/KpiRepository.kt
```diff
@@ -0,0 +1,57 @@
...
+   */
+  @Query(
+    //language=postgresql
+    """
+    -- this may include people no longer in prison
+    SELECT DISTINCT ON (prisoner_number) booking_id, prisoner_number
+    FROM prisoner_iep_level
+    JOIN next_review_date USING (booking_id)
+    WHERE next_review_date < CURRENT_DATE
+    ;
```
ushkarev, line 53: with `SELECT DISTINCT ON (…)` you need to have an `ORDER BY` clause so as to pick the latest review rather than _some_ review:

`ORDER BY prisoner_number, review_time DESC NULLS LAST`

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/Kpi.kt
xoen: I've copied the model from the others. So it should be the setup that we made it work.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonApiService.kt
xoen: This method has this boolean for consistency with the other methods in this class. But the `KpiService` will use the system token (by passing `true`)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/PrisonAPI.kt
xoen: Good spot. Even if we don't care for this field for the `KpiService` it could still cause deserialisation issues if some of the prisons have `null` long descriptions. I'll change 👌

xoen: Changed `Prison` dto class to have optional `longDescription`: https://github.com/ministryofjustice/hmpps-incentives-api/pull/503/commits/9fc99f7f11299240b9a2e7f330d7d9532f109b1b

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/Kpi.kt
ushkarev: the `IncentiveLevel`, for instance, has ` = LocalDateTime.now()`.

however `NextReviewDate` has ` = null` instead so at instantiation we have to do `whenUpdated = LocalDateTime.now(clock)`

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/KpiRepository.kt
xoen: Thanks for the explanation earlier Igor. I'm still not 100% sure the ORDER by would make a difference in our case (given we're only interested in the number of overdue prisoners) but I've added just to be on the safe side.

Also good to have a well-defined order given how `DISTINCT ON` work (e.g. it is a glorified group by).

I've also added `review_time` to the `SELECT`, even tho we don't use that data. But the order of the rows returned seem to change so I've deemed to be saver to add that too.

If the query still doesn't look right let me know: https://github.com/ministryofjustice/hmpps-incentives-api/pull/503/commits/aae18d5caca95a4dbdd3a3ef57e3c17e31945181

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/HmppsIncentivesApi.kt
```diff
@@ -10,6 +13,8 @@ import org.springframework.boot.runApplication
 const val SYSTEM_USERNAME = "INCENTIVES_API"
 
 @SpringBootApplication
+@EnableScheduling
+@EnableSchedulerLock(defaultLockAtMostFor = defaultLockAtLeastFor)
```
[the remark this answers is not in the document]
xoen: I was following along this article here: https://www.baeldung.com/shedlock-spring

But I wouldn't be surprised if these annotation worked wherever.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    println("Number of prisoners across the estate: ${allPrisoners.size}")
+
+    // Get prisoner numbers overdue, "filter" out people no longer in prison and return the count
+    kpiRepository
+      .getPrisonerNumbersOverdueReview()
+      .count { allPrisoners.contains(it.prisonerNumber) }
+  }
+
+  private fun getPrisons(): List<String> = runBlocking {
```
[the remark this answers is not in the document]
xoen: Mainly because I wanted to make all the other requests in concurrently.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/KpiRepository.kt
xoen: True. Although the little class may be more open for extension in the future. But I'll probably change.

ushkarev: honestly, i'm happy with a named object!

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/HmppsIncentivesApi.kt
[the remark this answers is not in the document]
ushkarev: cool beans! i was just checking!

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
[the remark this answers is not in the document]
ushkarev: cool

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/KpiRepository.kt
xoen: Hopefully this commit message make some sense 😅: [: Fix query to get overdue prisoners](https://github.com/ministryofjustice/hmpps-incentives-api/pull/503/commits/4896983ef9babbab0ecf9db0e2e493e5820bebef)

I tweaked the test so that the data had that edge case where a single prisoner number had multiple bookings, some with overdue review but some not overdue.

xoen: I've also simplified the return type of that repository method: 7ab3544fe5f049145caf12158079a9e42d0b5835

ushkarev (approved):
Worth a try, eh? 🤪

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/Kpi.kt
xoen: Just to recap our offline chat. Both fields are automatically set to `now()` correctly because of the DB table `DEFAULT now()` (similar to what happen for the next_review_date table/JPA entity).
If we were to update any of these records we'll have explicitly pass the `whenUpdated` field.

I don't think there is anything to do with this tbh.

PR #504, opened 2023-10-26
Use non-standard ports when using port forward tools

xoen:
Using the real port could lead to "accidents". For example if the gradle task to connect the preprod/prod DB is left running and tests are run at the same time they may potentially run against a non-test database which can cause all sort of problem (especially if the test then truncate the table)

It's better to err on the side of caution and use a non-standard port (`12345`) by default. This can be changed anyway for whatever reason.


xoen:
@ushkarev any objection to this?

ushkarev (approved):

PR #510, opened 2023-10-31
INC-1326 Add separate endpoints for add/read incentives - /iep/ will be deprecated

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/PrisonerIncentiveLevel.kt
```diff
@@ -5,12 +5,14 @@ import org.springframework.data.annotation.Id
...
 import org.springframework.data.domain.Persistable
+import org.springframework.data.relational.core.mapping.Table
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.IsRealReview
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.ReviewType
 import java.time.LocalDate
 import java.time.LocalDateTime
 
-data class PrisonerIepLevel(
+@Table("prisoner_iep_level")
+data class PrisonerIncentiveLevel(
```
xoen, line 15: I'd actually probably call this `IncentiveReview` if we were to review this.

The table contains the reviews and an instance of this is a single Incentive review.

It would also make most of the code that deals with review so more meaningful 'cause prisonerIepLevels have always confused me, especially as it almost sounds like a specific Incentive level rather than a review.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/jpa/repository/PrisonerIncentiveLevelRepository.kt
```diff
@@ -5,15 +5,15 @@ import org.springframework.data.r2dbc.repository.Modifying
...
+import uk.gov.justice.digital.hmpps.incentivesapi.jpa.PrisonerIncentiveLevel
 
 @Repository
-interface PrisonerIepLevelRepository : CoroutineCrudRepository<PrisonerIepLevel, Long> {
-  fun findAllByPrisonerNumberOrderByReviewTimeDesc(prisonerNumber: String): Flow<PrisonerIepLevel>
-  fun findAllByBookingIdOrderByReviewTimeDesc(bookingId: Long): Flow<PrisonerIepLevel>
-  fun findAllByBookingIdInAndCurrentIsTrueOrderByReviewTimeDesc(bookingIds: List<Long>): Flow<PrisonerIepLevel>
-  fun findAllByBookingIdInOrderByReviewTimeDesc(bookingIds: List<Long>): Flow<PrisonerIepLevel>
-  suspend fun findFirstByBookingIdOrderByReviewTimeDesc(bookingId: Long): PrisonerIepLevel?
+interface PrisonerIncentiveLevelRepository : CoroutineCrudRepository<PrisonerIncentiveLevel, Long> {
```
xoen, line 11: Similarly, I think this should be called `IncentiveReviewRepository`.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/ManageIncentiveReviewsResource.kt
```diff
@@ -0,0 +1,253 @@
...
+  )
+  suspend fun getPrisonerIncentiveReviewHistory(
+    @Schema(description = "Prisoner Number", example = "A1234AB", required = true, pattern = "^[A-Z0-9]{7}$")
+    @PathVariable
+    prisonerNumber: String,
+  ): IncentiveReviewSummary =
+    prisonerIncentiveReviewService.getPrisonerIncentiveHistory(prisonerNumber)
+
+  @PostMapping("/booking/{bookingId}")
+  @PreAuthorize("hasRole('MAINTAIN_IEP') and hasAuthority('SCOPE_write')")
```
xoen, line 173: IIRC this is the role users need to have right? But if these endpoint are meant to be used by system clients maybe we could introduce a separate role for this?

Maybe like `WRITE_INCENTIVES` or `WRITE_INCENTIVE_REVIEWS` similar to what we have in NONNA-API?

Also, you think the read role should be just called `READ_INCENTIVES` or `READ_INCENTIVE_REVIEWS`?
I'm probably in favour of being specific and call these `READ_INCENTIVE_REVIEWS`/`WRITE_INCENTIVE_REVIEWS` as it's more clear (especially as we shouldn't have any constraints on these names unless I'm missing something here)

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/ManageIncentiveReviewsResource.kt
```diff
@@ -0,0 +1,253 @@
...
+import uk.gov.justice.digital.hmpps.incentivesapi.dto.CreateIncentiveReviewRequest
+import uk.gov.justice.digital.hmpps.incentivesapi.dto.CurrentIncentiveLevel
+import uk.gov.justice.digital.hmpps.incentivesapi.dto.IncentiveReviewDetail
+import uk.gov.justice.digital.hmpps.incentivesapi.dto.IncentiveReviewSummary
+import uk.gov.justice.digital.hmpps.incentivesapi.service.PrisonerIncentiveReviewService
+import uk.gov.justice.digital.hmpps.incentivesapi.util.ensure
+
+@RestController
+@RequestMapping("/incentive-reviews", produces = [MediaType.APPLICATION_JSON_VALUE])
+@PreAuthorize("hasRole('ROLE_READ_INCENTIVES')")
```
xoen, line 30: As I mentioned [in this other comment](https://github.com/ministryofjustice/hmpps-incentives-api/pull/510/commits/813e741c3d845a38e5405939d1a0cea05a53fccd#r1378675627) I think it may be nicer to be more explicit and call this role `READ_INCENTIVE_REVIEWS` given:
- I think we have freedom here
- it's more explicit/clear

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/ManageIncentiveReviewsResource.kt
Mjwillis: we could use scope instead for read/write 
- INCENTIVE_REVIEWS with read and write scope

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IncentiveLevelReview.kt
```diff
@@ -4,7 +4,7 @@ import io.swagger.v3.oas.annotations.media.Schema
 import java.time.LocalDate
 
 @Schema(description = "Incentive review information for a prisoner")
-data class IncentiveReview(
+data class IncentiveLevelReview(
```
xoen, line 7: Why renaming to `IncentiveLevelReview` instead of the simpler `IncentiveReview`? Was it clashing with anything?

Mjwillis: yes the DTO

Mjwillis: was same as the JPA

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/ManageIncentiveReviewsResource.kt
xoen: We could, it kinda makes sense but I don't know if there is any best practice there or common way of doing this across the other APIs.
I know in NONNA-API we spelled it out in separate roles (maybe an argument for doing the same?) but I don't have any strong opinions.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IncentiveLevelReview.kt
xoen: Wouldn't they be in separate packages/namespaces and be fine tho?

Mjwillis: Yes but it will require further refactor, the package names will need to be explicit, however happy to do it

Mjwillis:
@xoen would you like to re-review?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IncentiveLevelReview.kt
Mjwillis: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IncentiveReviewSummary.kt
```diff
@@ -30,8 +30,8 @@ data class IepSummary(
   val iepTime: LocalDateTime,
   @Schema(description = "Location  of prisoner when review took place within prison (i.e. their cell)", example = "1-2-003", required = false)
   val locationId: String? = null,
-  @Schema(description = "IEP Review History (descending in time)", required = true)
-  var iepDetails: List<IepDetail>,
+  @Schema(description = "Incentive Review History (descending in time)", required = true)
+  var incentiveReviewDetails: List<IncentiveReviewDetail>,
```
ushkarev, line 34: changing the DTO’s field name here changes all existing apis, doesn’t it? i don’t think we want to do that…
(i like the improved description though)

src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/UnprotectedIncentiveReviewsResourceTestLevel.kt
```diff
@@ -72,13 +72,13 @@ class IepReviewsResourceTest : IncentiveLevelResourceTestBase() {
...
     webTestClient.post().uri("/iep/reviews/booking/$bookingId")
       .headers(setAuthorisation(roles = listOf("ROLE_MAINTAIN_IEP"), scopes = listOf("read", "write")))
-      .bodyValue(IepReview("STD", "A comment"))
+      .bodyValue(CreateIncentiveReviewRequest("STD", "A comment"))
       .exchange()
       .expectStatus().isCreated
 
     val today = now().format(DateTimeFormatter.ISO_DATE)
     val nextReviewDate = now().plusYears(1).format(DateTimeFormatter.ISO_DATE)
-    webTestClient.get().uri("/iep/reviews/booking/$bookingId?use-nomis-data=false")
```
ushkarev, line 81: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IncentiveReviewSummary.kt
Mjwillis: corrected

ushkarev (approved):

PR #515, opened 2023-11-03
INC-1327 Remove unused endpoint

Mjwillis:


ushkarev (approved):
are we totally sure it's not used? 🙉

PR #516, opened 2023-11-03
SDIT-1174 prison-api CaseNoteResource

steverendell:
call prison-api case notes with client token


ushkarev (approved):
@steverendell I was under the impression that throughout DPS there’s a move to use _system_ tokens when one api calls another…? No objection, obviously, if this is required! Do we need to add/remove any roles from the client used by this application?

Thoughts, @Mjwillis?

ushkarev:
wait i think i misunderstood

>call prison-api case notes with *client* token

because i've just seen the commit message says

> call case notes with *system* token, not user token

😮‍💨

PR #519, opened 2023-11-07
INC-1327 Remove unused endpoints

Mjwillis:
Just need to wait until a few services go live for this one:
Dependant on 
https://dsdmoj.atlassian.net/browse/HAAR-2027 
https://github.com/ministryofjustice/pathfinder/pull/1330 
https://github.com/ministryofjustice/hmpps-activities-management/pull/642
https://github.com/ministryofjustice/manage-soc-cases/pull/444


src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/resource/UnprotectedIncentiveReviewsResourceTestLevel.kt
```diff
@@ -1,173 +0,0 @@
-package uk.gov.justice.digital.hmpps.incentivesapi.resource
```
xoen, line 1: Would be better to only remove the tests for the specific endpoints that were removed instead of the whole test file for the resource?

Mjwillis: The endpoint is just a wrapper, doesn't seem worth keeping it as the final endpoint will go shortly

xoen (approved):

PR #523, opened 2023-11-14
Avoid calling `runBlocking` from within coroutines as recommended in documentation

ushkarev:
Ref: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/run-blocking.html

`runBlocking` blocks the current thread, but since coroutines share a small pool of threads (perhaps even 1 thread??) it means that this one cannot be used to progress any other coroutines. This could feasibly lead to a deadlock or, at the very least, defeats the purpose of coroutines.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
@@ -1,9 +1,9 @@
 package uk.gov.justice.digital.hmpps.incentivesapi.service
 
-import kotlinx.coroutines.async
-import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.asFlow
 import kotlinx.coroutines.flow.count
-import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.flow.map
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.OffenderSearchPrisoner
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.prisonapi.Prison
@@ -18,31 +18,33 @@ class KpiService(
   private val offenderSearchService: OffenderSearchService,
 ) {
 
-  fun getNumberOfReviewsConductedAndPrisonersReviewed(day: LocalDate): ReviewsConductedPrisonersReviewed = runBlocking {
-    kpiRepository.getNumberOfReviewsConductedAndPrisonersReviewed(day)
+  suspend fun getNumberOfReviewsConductedAndPrisonersReviewed(day: LocalDate): ReviewsConductedPrisonersReviewed {
+    return kpiRepository.getNumberOfReviewsConductedAndPrisonersReviewed(day)
   }
 
-  fun getNumberOfPrisonersOverdue(): Int = runBlocking {
+  suspend fun getNumberOfPrisonersOverdue(): Int {
     // Get a set containing all the prisonerNumber of everyone in prison (any prison)
     val allPrisoners = mutableSetOf<String>()
-    val prisonersInPrisons = getPrisons().map { prisonId -> async { getPrisonersInPrison(prisonId) } }.awaitAll()
-    prisonersInPrisons.forEach { prisoners ->
+    getPrisons().map { prisonId ->
+      getPrisonersInPrison(prisonId)
+    }.collect { prisoners ->
       allPrisoners.addAll(prisoners.map(OffenderSearchPrisoner::prisonerNumber))
     }
 
     println("Number of prisoners across the estate: ${allPrisoners.size}")
 
     // Get prisoner numbers overdue, "filter" out people no longer in prison and return the count
-    kpiRepository
+    return kpiRepository
       .getPrisonerNumbersOverdueReview()
       .count { allPrisoners.contains(it) }
   }
 
-  private fun getPrisons(): List<String> = runBlocking {
+  private suspend fun getPrisons(): Flow<String> {
     println("Getting list of prisons from Prison API...")
-    prisonApiService
+    return prisonApiService
       .getActivePrisons(true)
       .map(Prison::agencyId)
+      .asFlow()
   }
 
   private suspend fun getPrisonersInPrison(prisonId: String): List<OffenderSearchPrisoner> {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
@@ -1,9 +1,9 @@
 package uk.gov.justice.digital.hmpps.incentivesapi.service
 
-import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.asFlow
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.count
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/NextReviewDateUpdaterService.kt
```diff
@@ -93,24 +91,22 @@ class NextReviewDateUpdaterService(
     return nextReviewDatesAfterUpdate
   }
 
-  private suspend fun publishDomainEvents(
+  private fun publishDomainEvents(
```
ushkarev, line 94: `publishDomainEvents` was called from inside a suspend function and used `runBlocking` – this is not recommended. Since there are no suspend function calls inside `publishDomainEvents`, special coroutine handling can be removed as it likely added unnecessary overhead anyway.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/task/UpdateKpis.kt
```diff
@@ -28,13 +28,13 @@ class UpdateKpis(
     lockAtLeastFor = defaultLockAtLeastFor,
     lockAtMostFor = defaultLockAtMostFor,
   )
-  fun updateKpis() {
+  fun updateKpis(): Unit = runBlocking {
```
ushkarev, line 31: This is a top-level non-suspend function so `runBlocking` can be lifted out to the top of it to reduce overhead of creating multiple blocking scopes

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
```diff
@@ -18,31 +18,33 @@ class KpiService(
   private val offenderSearchService: OffenderSearchService,
 ) {
 
-  fun getNumberOfReviewsConductedAndPrisonersReviewed(day: LocalDate): ReviewsConductedPrisonersReviewed = runBlocking {
```
ushkarev, line 21: `getNumberOfReviewsConductedAndPrisonersReviewed` was called from inside a coroutine so should avoid using `runBlocking`

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
```diff
@@ -18,31 +18,33 @@ class KpiService(
...
-    kpiRepository.getNumberOfReviewsConductedAndPrisonersReviewed(day)
+  suspend fun getNumberOfReviewsConductedAndPrisonersReviewed(day: LocalDate): ReviewsConductedPrisonersReviewed {
+    return kpiRepository.getNumberOfReviewsConductedAndPrisonersReviewed(day)
   }
 
-  fun getNumberOfPrisonersOverdue(): Int = runBlocking {
```
ushkarev, line 25: `getNumberOfPrisonersOverdue` was called from inside a coroutine so should avoid using `runBlocking`

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
```diff
@@ -18,31 +18,33 @@ class KpiService(
...
+  suspend fun getNumberOfPrisonersOverdue(): Int {
     // Get a set containing all the prisonerNumber of everyone in prison (any prison)
     val allPrisoners = mutableSetOf<String>()
-    val prisonersInPrisons = getPrisons().map { prisonId -> async { getPrisonersInPrison(prisonId) } }.awaitAll()
-    prisonersInPrisons.forEach { prisoners ->
+    getPrisons().map { prisonId ->
+      getPrisonersInPrison(prisonId)
+    }.collect { prisoners ->
```
ushkarev, line 30: Turning `getPrisons()` into a Flow means that the suspend function `getPrisonersInPrison` can be called on each item without launching deferred jobs using `async`

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
```diff
@@ -18,31 +18,33 @@ class KpiService(
...
 
     // Get prisoner numbers overdue, "filter" out people no longer in prison and return the count
-    kpiRepository
+    return kpiRepository
       .getPrisonerNumbersOverdueReview()
       .count { allPrisoners.contains(it) }
   }
 
-  private fun getPrisons(): List<String> = runBlocking {
+  private suspend fun getPrisons(): Flow<String> {
```
ushkarev, line 42: `getPrisons` was called from inside a coroutine so should avoid using `runBlocking`

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
ushkarev: I don’t think we need a wrapping `coroutineScope` to handle anything since a `Flow` deals with cancellation already

xoen: Hang on, but why not launch the coroutines? Wouldn't this make the requests to get the prisoners in each prison _**sequential**_, rather than concurrently?

These requests should be independent and it would be best to do them all "at the same time" no?

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/task/UpdateKpis.kt
```diff
@@ -28,13 +28,13 @@ class UpdateKpis(
...
     val day = LocalDate.now()
 
-    val kpiExists = runBlocking { kpiRepository.existsById(day) }
+    val kpiExists = kpiRepository.existsById(day)
     if (kpiExists) {
       LOG.debug("KPI for $day already exists, skipping...")
-      return
+      return@runBlocking
```
xoen, line 37: What's this doing? Looks like a label jump, but what's the difference with a plain `return` in this case?

ushkarev: it's just what you have to do apparently; intellij makes you do this. `return` would exit the function, but `return@runBlocking` exits the "block" rather than the whole function

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
ushkarev: yeah you're very likely right! because Flows are cold, it's likely that `getPrisonersInPrison` would only be called on each item after the previous one is done

REVERTED

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/task/UpdateKpis.kt
xoen: I'm not going to go against IntelliJ then 😅 ... as long as the behaviour didn't change and this return skips the INSERT 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/KpiService.kt
xoen: I mean this all happens in the background and hopefully it would still be quick enough to complete within the task allowed timeframe...but it feels right to make these requests concurrently as they're all independent.

xoen (approved):

xoen (approved):

Mjwillis (approved):
LGTM

PR #546, opened 2024-03-08
Update all non major Gradle dependencies

Mjwillis:


ushkarev:
renovate wants to take opentelemetry-instrumentation-annotations to a higher version in #541 – do you want to keep it at 1.x?

Mjwillis:
I really don't know - feels wrong

ushkarev (approved):

ushkarev:
interestingly, `io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations` 2.1.0 only depends on the 1.35.0 version of `io.opentelemetry:opentelemetry-api` according to [maven](https://mvnrepository.com/artifact/io.opentelemetry.instrumentation/opentelemetry-instrumentation-annotations/2.1.0).
so it doesn’t look like they align major versions :thinking_face:

also, gradle doesn’t list multiple versions of any opentelemetry library so i guess the dependency web is self-consistent

renovate has already upgraded NON and IRS this combo… and they seem to run ok?

PR #547, opened 2024-03-11
INC-1355 Add SubjectAccessRequestService and update hmpps-kotlin-spring-boot-starter

Mjwillis:
This commit introduces the SubjectAccessRequestService in src/main which is used to handle subject access requests. Also, 'uk.gov.justice.service.hmpps:hmpps-kotlin-spring-boot-starter' in build.gradle.kts has been updated to version 0.2.2. Additionally, SARResourceTest is added into src/test to provide unit tests for this new service.


petergphillips (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/SubjectAccessRequestService.kt
```diff
@@ -0,0 +1,71 @@
...
+  private val incentiveReviewRepository: IncentiveReviewRepository,
+  private val nextReviewRepository: NextReviewDateRepository,
+) : HmppsPrisonSubjectAccessRequestReactiveService {
+
+  override suspend fun getPrisonContentFor(
+    prn: String,
+    fromDate: LocalDate?,
+    toDate: LocalDate?,
+  ): HmppsSubjectAccessRequestContent? {
+    val reviews = incentiveReviewRepository.findAllByPrisonerNumberOrderByReviewTimeDesc(prn)
```
ushkarev, line 26: i think it’d be better to use the `PrisonerIncentiveReviewService.getPrisonerIncentiveHistory` service method which already sorts out calculating the next review date if it is not available in the repository…

Mjwillis: Yes true, was just trying to get the raw data but agree

ushkarev (dismissed):

ushkarev (approved):

Mjwillis (commented):
Lgtm

PR #551, opened 2024-03-14
INC-1356 Update UUID for queue names and simplify queue tests

Mjwillis:
This update changes hard-coded queue names and ARNs in YAML files to generated UUIDs, improving flexibility for testing and development. Furthermore, the complexity of tests in IncentiveLevelResourceTestBase has been reduced by removing extra setup steps for domain events subscription, thanks to adding "test" queue in the SqsIntegrationTestBase and the application-test.yml files.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/integration/SqsIntegrationTestBase.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/integration/SqsIntegrationTestBase.kt
@@ -42,6 +42,7 @@ class SqsIntegrationTestBase : IntegrationTestBase() {
 
   protected val auditQueue by lazy { hmppsQueueService.findByQueueId("audit") as HmppsQueue }
   protected val incentivesQueue by lazy { hmppsQueueService.findByQueueId("incentives") as HmppsQueue }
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/integration/SqsIntegrationTestBase.kt
```diff
@@ -42,6 +42,7 @@ class SqsIntegrationTestBase : IntegrationTestBase() {
 
   protected val auditQueue by lazy { hmppsQueueService.findByQueueId("audit") as HmppsQueue }
   protected val incentivesQueue by lazy { hmppsQueueService.findByQueueId("incentives") as HmppsQueue }
```
ushkarev, line 44: if i'm reading this correctly, we can remove `incentivesQueue` and use `testDomainEventQueue` in its place in tests? because ever usage of this property just gets the SQS client but then reads messages from the new `testDomainEventQueue.queueUrl`

Mjwillis: yes

ushkarev (approved):

PR #581, opened 2024-06-03
INC-1367: Use generic-service's DLQ CronJob

xoen:
We currently have a "custom" CronJob which retries all the messages in the DLQ (Dead Letter Queue).

However the "same" CronJob is provided by the `generic-service` helm chart we use to deploy the service.
This CronJob has the advantae of being aware of the scheduled downtime and avoiding trying to run this cronjob when the service is not running (as that would not work).

This may (or may not) reduce the number of failures/alerts we see in the DPS alert channels related to this cronjob failing.


helm_deploy/hmpps-incentives-api/values.yaml
```diff
@@ -12,6 +12,9 @@ generic-service:
     tag: app_version    # override at deployment time
     port: 8080
 
+  retryDlqCronjob:
+    enabled: true
```
xoen, line 16: This setting creates a CronJob very similar (aside from the `schedule`) to the one we had before: https://github.com/ministryofjustice/hmpps-helm-charts/blob/85b24e1789449cf3306fe9229cf6dbef088d9ae2/charts/generic-service/templates/retry-dlq-cronjob.yaml

ushkarev (approved):

helm_deploy/hmpps-incentives-api/values.yaml
xoen: When the scheduled downtime is enabled (as it is for `dev` and `preprod` in our case) this is the schedule used for this CronJob:

```
# every 10 minutes between 7am and 10pm UTC Monday-Friday
retryDlqSchedule: "*/10 7-21 * * 1-5"
```

https://github.com/ministryofjustice/hmpps-helm-charts/blob/85b24e1789449cf3306fe9229cf6dbef088d9ae2/charts/generic-service/values.yaml#L237

PR #586, opened 2024-06-19
INC-1335: Fix async-api.yml "duplicated" keys

xoen:


xoen:
Not needed, fixed in https://github.com/ministryofjustice/hmpps-domain-events/pull/30

PR #592, opened 2024-07-08
INC-1335: Sync changes from templates

ushkarev:
Ref:
- [hmpps-ip-allowlists#21](https://github.com/ministryofjustice/hmpps-ip-allowlists/pull/21)
- [hmpps-template-kotlin#215](https://github.com/ministryofjustice/hmpps-template-kotlin/pull/215)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/AuditService.kt
```diff
@@ -32,7 +33,7 @@ class AuditService(
   suspend fun sendMessage(auditType: AuditType, id: String, details: Any, username: String? = null) {
     val auditEvent = AuditEvent(
       what = auditType.name,
-      who = username ?: authenticationFacade.getUsername(),
+      who = username ?: authenticationHolder.getUsername() ?: SYSTEM_USERNAME,
```
petergphillips, line 36: looking at your existing facade implementation this looks like it would return the `clientId` if the `username` isn't set.  The equivalent would be to call `getPrincipal` instead, would be interesting to see what data is in the audit events for the `username` field at present

ushkarev: poop. good spot!

Mjwillis (approved):

PR #596, opened 2024-07-23
INC-1371 (part 1): Stop populating locationId in Incentives reviews

xoen:
This information shouldn't be used anywhere and it's nullable.

Historically it was added to the new service thinking "it may be interesting/useful" but it never got used and Incentives is the wrong place to store location information anyway.

This means avoiding a bunch of requests to the deprecated Prison API's endpoint `GET /api/locations/{locationId}` (this is one of the endpoint actively being pushed out and replaced by new Locations API).

It also allows us to simplify codebase a little bit.

Existing `locationId` data in the DB is returned as before, at least for now. In the future we will remove the DB column and deprecated fields.


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIncentiveReviewService.kt
```diff
@@ -176,26 +176,16 @@ class PrisonerIncentiveReviewService(
       val iepLevel = getIncentiveLevelForReviewType(prisonerInfo, reviewType)
       val comment = getReviewCommentForEvent(prisonOffenderEvent)
 
-      val createIncentiveReviewRequest = CreateIncentiveReviewRequest(
-        iepLevel = iepLevel,
-        comment = comment,
-        reviewType = reviewType,
-      )
-
-      val locationInfo = prisonApiService.getLocationById(prisonerInfo.assignedLivingUnitId, true)
```
xoen, line 185: This is using the client credentials which may explain why in AppInsights some requests seems to come from the UI ([see tracking spreadsheet](https://docs.google.com/spreadsheets/d/1xa0MJNMreQcqONtS9alkF_5Hg3NvWSCckDIyr2KQOXA/edit?gid=0#gid=0)).

jnioche-jd (approved):
LGTM

PR #598, opened 2024-07-24
INC-1371 (part 2): Get location description from Locations API

xoen:
⚠️ ~This PR is on a branch off [this other one](https://github.com/ministryofjustice/hmpps-incentives-api/pull/596)~ ⚠️  

The [reviews endpoint](https://incentives-api-dev.hmpps.service.justice.gov.uk/webjars/swagger-ui/index.html?configUrl=/v3/api-docs#/Incentive%20review%20summary/getReviews)'s response has a `locationDescription` field with the human description of a location (e.g. Houseblock 1). This is displayed in the UI reviews table page (this endpoint is "internal" and used only by the UI).

Before this PR this description was populated using the now deprecated `GET /api/locations/code/{locationId}` in Prison API endpoint.

The service now makes a request to the new Locations Inside Prison API. Specifically it uses the [`GET /locations/key/{key}` endpoint](https://locations-inside-prison-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/Locations/getLocationByKey) which takes a location "key" (e.g. `MDI-1`) and returns information about the location.

In our case we're interested in the `localName` field. This field is optional - presumably because it's not populated for individual cells? - hence the fallback to `pathHierarchy` field.

#### Notes on permissions
- The requests to the Locations API are made using the system client rather than the user/client token. So individual users of the UI don't need any special roles (similarly to what we do when getting information from Offender Search API)
  - Previously the information about a location were retrieved using the client token. This probably explains why [upstream API sees these requests as made from the UI itself](https://docs.google.com/spreadsheets/d/1xa0MJNMreQcqONtS9alkF_5Hg3NvWSCckDIyr2KQOXA/edit?gid=0#gid=0).
- This endpoint requires the `VIEW_LOCATIONS` role. Mike kindly already added this to the `dev`/`preprod`/`prod` clients (but it may be worth triple check with someone)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -69,6 +69,16 @@ services:
     volumes:
       - ./prisoner-search-data:/home/wiremock
 
+  locations-inside-prison:
+    image: wiremock/wiremock:2.35.0
+    networks:
+      - hmpps
+    container_name: locations-inside-prison
+    ports:
+      - "8085:8080"
+    volumes:
+      - ./locations-inside-prison-data:/home/wiremock
+
   prison-api:
     image: quay.io/hmpps/prison-api:latest
     networks:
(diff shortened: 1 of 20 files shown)


jnioche-jd:
> In our case we're interested in the localName field. This field is optional - presumably because it's not populated for individual cells? - hence the fallback to pathHierarchy field.

Exactly, see 
https://github.com/ministryofjustice/hmpps-locations-inside-prison-api/blob/aaded2ed9c3647a34f5eb5f4fdf3185a90fd7f62/src/main/kotlin/uk/gov/justice/digital/hmpps/locationsinsideprison/dto/LocationDto.kt#L49

jnioche-jd (dismissed):
Looks good. Compiles fine and tests OK

PR #669, opened 2025-05-20
IR-1053 Update Kotlin and SonarQube plugins, and upgrade OpenTelemetry instrumentation dependency

Mjwillis:


xoen (approved):

build.gradle.kts
```diff
@@ -53,7 +53,7 @@ dependencies {
   implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
   implementation("com.fasterxml.jackson.datatype:jackson-datatype-jsr310")
   implementation("io.opentelemetry:opentelemetry-api:1.50.0")
-  implementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:2.15.0")
+  implementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:2.16.0")
```
ushkarev, line 56: as in the other 2 api repos, renovate also says in #667 to update `kotlin("jvm") version "2.1.21"` in the other build.gradle.kts file

i hadn't looked at the errors so you may be right about ignoring `testImplementation("io.swagger.parser.v3:swagger-parser:2.1.28")`!

the only problem **that i know of** with renovate's automatic prs is that 
`testRuntimeOnly("org.junit.platform:junit-platform-launcher:1.11.4")`
must be kept at that version

ushkarev (approved):

PR #683, opened 2025-07-17
INC-1395: Use Prisoner Search API to get prisoner info

xoen:
~Two ~Prisoner Search API is now used to get prisoner information by `prisonerNumber`:
- [`GET /prisoner/{prisonerNumber}`](https://prisoner-search-dev.prison.service.justice.gov.uk/swagger-ui/index.html?configUrl=/v3/api-docs/swagger-config#/Popular/findByPrisonNumber) to get info by prisoner number
- ~[`POST /prisoner-search/booking-ids`](https://prisoner-search-dev.prison.service.justice.gov.uk/swagger-ui/index.html?configUrl=/v3/api-docs/swagger-config#/Batch/findByIds) to get info by `bookingId`~

A couple of field names are different (`agencyId` => `prisonId` / `offenderNo` => `prisonerNumber`) but the response from Prisoner Search API should be equivalent and much faster to get.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/PrisonAPI.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/PrisonAPI.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.incentivesapi.dto.prisonapi
 
 import org.springframework.format.annotation.DateTimeFormat
 import uk.gov.justice.digital.hmpps.incentivesapi.dto.PrisonerAlert
+import uk.gov.justice.digital.hmpps.incentivesapi.dto.PrisonerBasicInfo
 import uk.gov.justice.digital.hmpps.incentivesapi.service.PrisonerInfoForNextReviewDate
 import java.time.LocalDate
 import java.time.LocalDateTime
@@ -31,12 +32,6 @@ data class BookingFromDatePair(
   @param:DateTimeFormat(iso = DateTimeFormat.ISO.DATE_TIME)
   val fromDate: LocalDateTime,
 )
-
-interface PrisonerBasicInfo {
-  val bookingId: Long
-  val prisonerNumber: String
-  val prisonId: String
-}
 
 data class PrisonerInfo(
   override val bookingId: Long,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/PrisonAPI.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/PrisonAPI.kt
@@ -1,8 +1,6 @@
 package uk.gov.justice.digital.hmpps.incentivesapi.dto.prisonapi
 
 import org.springframework.format.annotation.DateTimeFormat
-import uk.gov.justice.digital.hmpps.incentivesapi.service.PrisonerInfoForNextReviewDate
-import java.time.LocalDate
 import java.time.LocalDateTime
 
 data class Prison(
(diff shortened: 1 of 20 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/PrisonAPI.kt
```diff
@@ -13,39 +11,6 @@ data class Prison(
   val active: Boolean,
 )
 
-data class PrisonerInfo(
-  val bookingId: Long,
-  val facialImageId: Long,
-  val firstName: String,
-  val lastName: String,
-  val offenderNo: String,
-  val agencyId: String,
```
xoen, line 22: This is `prisonId` in Prisoner Search API

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/prisonapi/PrisonAPI.kt
```diff
@@ -13,39 +11,6 @@ data class Prison(
...
-)
-
-data class PrisonerAlert(
```
xoen, line 25: Moved under PrisonerSearch dto.

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerSearchService.kt
```diff
@@ -64,4 +65,52 @@ class PrisonerSearchService(
...
+  }
+
+  /**
+   * Get a prisoner info by bookingId
+   * Requires role GLOBAL_SEARCH or PRISONER_SEARCH
+   */
+  suspend fun getPrisonerInfo(bookingId: Long): Prisoner {
+    val results = prisonerSearchWebClient.post()
+      .uri(
+        "/prisoner-search/booking-ids?" +
```
xoen, line 94: NOTE: There isn't a way to get a single prisoner by bookingId but there is an endpoint to get multiple, so using that but to get information by passing a single `bookingId` in the request body.

ushkarev (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerSearchService.kt
Mjwillis: We need to check that it can handle booking Id - Not sure let me check because search only hold the LATEST booking ID not older ones.  I need to check WHEN it looks up by bookingID 

If not we may still have to use prisonId for this endpoint

Mjwillis (commented):
See comment about booking Id

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIncentiveReviewService.kt
```diff
@@ -219,8 +220,8 @@ class PrisonerIncentiveReviewService(
     }
   }
 
-  private suspend fun getIncentiveLevelForReviewType(prisonerInfo: PrisonerInfo, reviewType: ReviewType): String {
-    val prisonIncentiveLevels = prisonIncentiveLevelService.getActivePrisonIncentiveLevels(prisonerInfo.agencyId)
+  private suspend fun getIncentiveLevelForReviewType(prisonerInfo: PrisonerBasicInfo, reviewType: ReviewType): String {
```
xoen, line 223: Technically I didn't _have to_ change the signature of this method as it's always invoked with the response from Prisoner Search API now (as the SNS domain event has the prisoner number)...

...but the method body only use basic information so it made sense to restrict the input to the things the method uses (it only uses `prisonId` and `bookingId` properties)

PR #757, opened 2026-04-28
IR-1572 Upgrade Spring Boot to v10 (hmpps-gradle-spring-boot 10.2.3) 

Mjwillis:
## Summary

Upgrades the HMPPS Gradle Spring Boot plugin from `9.3.0` to `10.2.3`, which brings in a new major version of the hmpps-kotlin-spring-boot-starter (`1.8.2` → `2.1.1`) and its associated breaking changes. This required removing hand-rolled auth/tracking infrastructure that is now provided by the starter, updating test infrastructure for new Spring Boot 3.x module paths, and refactoring a computed DTO property to make it clock-injectable and testable.

---

## Dependency upgrades

| Dependency | Before | After |
|---|---|---|
| `hmpps-gradle-spring-boot` plugin | `9.3.0` | `10.2.3` |
| `hmpps-kotlin-spring-boot-starter` | `1.8.2` | `2.1.1` |
| `hmpps-kotlin-spring-boot-starter-test` | `1.8.2` | `2.1.0` |
| `hmpps-sqs-spring-boot-starter` | `5.6.3` | `7.3.1` |
| `springdoc-openapi-starter-webflux-ui` | `2.8.15` | `3.0.3` |
| `kotlin` plugin | `2.3.0` | `2.3.21` |
| `kotlinx-coroutines` | implicit | `1.10.2` (explicit) |
| `shedlock` | `7.6.0` | `7.7.0` |
| `opentelemetry-instrumentation-annotations` | `2.24.0` | `2.27.0` |
| `opentelemetry-api` | `1.58.0` | replaced with `opentelemetry-extension-kotlin:1.61.0` |
| `org.sonarqube` plugin | `7.2.2.6593` | `7.2.3.7755` |
| `testcontainers` localstack/postgresql | `org.testcontainers:...:1.21.4` | `org.testcontainers:testcontainers-...:2.0.4` |
| Java/JVM target | 24 | 25 |

New explicit dependencies required by the new starter version:

- `spring-boot-starter-webclient`
- `spring-boot-starter-security` + `security-oauth2-client` + `spring-security-access`
- `spring-boot-starter-flyway`
- PostgreSQL drivers pinned to explicit versions (`r2dbc-postgresql:1.1.1.RELEASE`, `postgresql:42.7.10`)

---

## Infrastructure removed (now provided by starter)

Three hand-rolled auth/tracking classes have been deleted because `hmpps-kotlin-spring-boot-starter:2.x` now provides equivalent functionality via autoconfiguration:

- **`AuthHeaderWebFilter`** — propagated the `Authorization` header into the Reactor context. Now handled by the starter.
- **`AuthTokenFilterFunction`** — read the token back out of the Reactor context and attached it to outbound `WebClient` requests. Replaced by `reactiveUsernameAwareTokenRequestOAuth2AuthorizedClientManager` from the starter.
- **`OAuth2ClientConfiguration`** — manually configured a `ReactiveOAuth2AuthorizedClientManager` using `ReactiveGlobalPrincipalOAuth2AuthorizedClientService`. This is now autoconfigured by the starter.
- **`ClientTrackingWebFilter`** + its test — parsed the JWT to set OpenTelemetry span attributes. Now provided by the starter.

---

## `WebClientConfiguration` refactor

- `prisonWebClient` (the user-context-passing variant) now uses `reactiveUsernameAwareTokenRequestOAuth2AuthorizedClientManager` instead of the deleted `AuthTokenFilterFunction`.
- Added separate health web clients for `prison`, `locations`, and `prisonerSearch` (previously `prisonWebClient` was overloaded for health checks).
- Fixed a bug where `healthTimeout` was mistakenly used for all web client request timeouts — now `healthTimeout` (`10s`) is only used for health pings and `timeout` (`2m`) is used for API calls.
- Renamed `authWebClient` bean → `authHealthWebClient` to match its health-check-only purpose (and fixed `HealthPingCheck` qualifier accordingly).
- Replaced the hardcoded `api.health-timeout-ms` property with `api.health-timeout` (Duration) and `api.timeout` (Duration).

---

## OpenAPI configuration simplified

`OpenApiConfiguration` is significantly reduced following the springdoc 3.x update:

- Removed the manual OAuth2 flow registration (clientCredentials + authorizationCode) — springdoc 3.x handles this differently.
- Removed the `openAPICustomiser` bean that replaced `DateTimeSchema` properties with a custom `StringSchema` — no longer needed.
- Removed `License` from the API info.
- Removed the `oauthUrl` constructor parameter.
- Added a description tag for the `hmpps-queue-resource-async` endpoints.
- Security scheme simplified to a single `bearer-jwt` definition.

---

## DTO / service refactor

`IncentiveReviewSummary.daysSinceReview` was previously a computed `get()` property backed by `Clock.systemDefaultZone()`, which made it impossible to test deterministically without mocking the system clock. It has been refactored to:

1. A plain constructor parameter `val daysSinceReview: Int` on the data class.
2. A top-level function `daysSinceReviewCalc(iepDate, clock)` in the same file that callers use to compute the value before construction.

`PrisonerIncentiveReviewService` now passes a `clock` dependency to compute `daysSinceReview` at the point of building the response, and the unused `prisonApiService` injection was removed from the service.

---

## Test refactoring

Required to maintain compatibility with the new Spring Boot module structure:

- **`@DataR2dbcTest` import path** changed across all 5 repository tests:
    - Before: `org.springframework.boot.test.autoconfigure.data.r2dbc.DataR2dbcTest`
    - After: `org.springframework.boot.data.r2dbc.test.autoconfigure.DataR2dbcTest`

- **`IntegrationTestBase`**: Added `@AutoConfigureWebTestClient` — required by the new test autoconfiguration structure in the updated test starters.

- **`TestBase`**: Added shared `clock`, `now`, and `FIXED_TIME` constants so repository and integration tests share a single deterministic time reference rather than each test defining its own.

- **`IncentiveReviewSummaryDateTest`**: Updated to call the new top-level `daysSinceReviewCalc` function instead of the removed `incentiveReviewSummary.daysSinceReviewCalc(clock)` method.

- **`PrisonerIncentiveReviewServiceTest`**: Removed the now-unused `prisonApiService` mock.

- Added explicit test starter dependencies to align with the new modular Spring Boot test setup:
  `spring-boot-starter-data-r2dbc-test`, `spring-boot-starter-security-test`, `spring-boot-starter-webclient-test`, `spring-boot-starter-webflux-test`.

---

## Configuration changes (`application.yml`)

- Jackson date serialisation config removed — now handled by the starter's autoconfiguration.
- `spring.data.web.pageable.max-page-size` moved to correct YAML nesting level (was incorrectly nested under `spring.devtools.data` due to indentation error).
- `spring.devtools.*` config removed (devtools dependency also removed from build).
- `server.error.include-message: always` removed (the starter now controls error message exposure appropriately).
- `api.health-timeout-ms` replaced with `api.health-timeout: 10s` and `api.timeout: 2m`.

---

## Exception handler cleanup

Nullable `ResponseEntity<ErrorResponse?>?` return types tightened to `ResponseEntity<ErrorResponse>?` across all handler methods in `HmppsIncentivesApiExceptionHandler` — the inner nullability was unnecessary and flagged by the Kotlin compiler with the newer version.


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/IncentiveReviewSummary.kt
```diff
@@ -50,21 +49,14 @@ data class IncentiveReviewSummary(
...
-      return daysSinceReviewCalc(Clock.systemDefaultZone())
-    }
+  @param:Schema(description = "Days since last review", example = "23", required = true)
+  val daysSinceReview: Int,
+)
 
-  @JsonIgnore
-  fun daysSinceReviewCalc(clock: Clock): Int {
-    val today = LocalDate.now(clock).atStartOfDay()
-    return Duration.between(iepDate.atStartOfDay(), today).toDays().toInt()
```
xoen, line 66: I'm probably missing something but where was `iepDate` coming from before?

xoen: Ah, it was a method on the `IncentiveReviewSummary` class 👌

xoen (approved):

PR #767, opened 2026-05-06
IR-1612: Replace Prison API case note count usage with Case Notes API

Mjwillis:
## Background

The incentives service calculates positive and negative behaviour tallies for each prisoner by counting case notes of specific types since their last incentive review. This data is displayed on the incentive reviews screen and drives the `positiveBehaviours` / `negativeBehaviours` fields in the review response.

Previously this was fetched from **Prison API** via `POST /api/case-notes/usage-by-types`, which accepted a list of `(bookingId, fromDate)` pairs — one entry per prisoner, each with their own personalised start date — and returned a flat list of counts.

That endpoint is being deprecated. This PR migrates the same functionality to the **Case Notes API** (`POST /case-notes/usage`), which has a materially different contract:

| Concern | Prison API (old) | Case Notes API (new) |
|---|---|---|
| Prisoner identifier | `bookingId: Long` | `personIdentifier: String` (prison number) |
| Per-prisoner start dates | Yes — one `(bookingId, fromDate)` per prisoner | No — a single `from` date applies to all prisoners in the request |
| Batching | All prisoners in one call | One call per unique `from` date |
| Response shape | `List<PrisonerCaseNoteByTypeSubType>` (flat) | `{ content: Map<personIdentifier, List<UsageResponse>> }` |

---

### 1. New `CaseNotesApiService`

`src/main/kotlin/.../service/CaseNotesApiService.kt` _(new file)_

A new Spring service wrapping the Case Notes API client. The single method `retrieveCaseNoteCountsByFromDate` posts a `UsageByPersonIdentifierRequest` to `/case-notes/usage` and streams the response back as a `Flow<NoteUsageResponse>`.

New DTOs introduced alongside the service (all in `CaseNotesApiService.kt`):

- **`UsageByPersonIdentifierRequest`** — request body: a set of `typeSubTypes`, a set of `personIdentifiers`, and a single `from` date
- **`TypeSubTypeRequest`** — a case note type with an optional set of sub-types
- **`NoteUsageResponse`** — response wrapper: `content: Map<String, List<UsageByPersonIdentifierResponse>>` keyed by `personIdentifier`
- **`UsageByPersonIdentifierResponse`** — one entry per prisoner/type/subType combination: `personIdentifier`, `type`, `subType`, `count`, `latestNote?`
- **`LatestNote`** — timestamp of the most recent matching note

---

### 2. `BehaviourService` refactored

`src/main/kotlin/.../service/BehaviourService.kt`

**Dependency change:** `PrisonApiService` replaced by `CaseNotesApiService`.

**Identifier mapping:** Because the new API uses prison number (`personIdentifier`) but downstream code keys everything off `bookingId`, two lookup maps are built from the incoming reviews:

- `personIdentifierToBookingId: Map<String, Long>` — used when mapping API responses back to `PrisonerCaseNoteByTypeSubType`
- `bookingIdToPersonIdentifier: Map<Long, String>` — used when constructing API requests

**API call strategy — batching by `from` date:**

The new API can only accept a single `from` date per call, so a separate call is required for each distinct start date. However, since incentive reviews are conducted roughly annually and case notes are only counted over the preceding 3 months, many prisoners will share the same effective start date (the 3-month default). To avoid unnecessary calls, prisoners are **grouped by their truncated review date** before calling the API:

```
lastReviewsOrDefaultPeriods.entries
  .groupBy({ it.value }) { it.key }           // Map<LocalDateTime, List<bookingId>>
  .map { (fromDate, bookingIds) ->
      retrieveCaseNoteCountsByFromDate(
          typeSubTypes,
          bookingIds.mapToPersonIdentifiers(),  // Set<String>
          fromDate,
      )
  }
```

This means:
- Prisoners whose last real review falls within the 3-month window each get their own call (different dates)
- Prisoners whose last real review predates 3 months, or who have never had a real review, all share the same 3-month default date and are **batched into a single call**

In the common steady-state scenario where most prisoners had their last review at roughly the same time, this can reduce the number of API calls significantly.

**`transformCaseNoteResults` (new private method):**

Collects all `Flow<NoteUsageResponse>` objects emitted from the per-date calls, then flattens each response's `content` map into a list of `PrisonerCaseNoteByTypeSubType`, resolving `personIdentifier → bookingId` via the lookup map:

```kotlin
private suspend fun transformCaseNoteResults(
    result: List<Flow<NoteUsageResponse>>,
    personIdentifierToBookingId: Map<String, Long>,
): List<PrisonerCaseNoteByTypeSubType> =
    result.flatMap { it.toList() }
        .flatMap { response ->
            response.content.flatMap { (personIdentifier, usages) ->
                usages.map { usage ->
                    PrisonerCaseNoteByTypeSubType(
                        bookingId = personIdentifierToBookingId[personIdentifier] ?: error(...),
                        caseNoteType = usage.type,
                        caseNoteSubType = usage.subType,
                        numCaseNotes = usage.count,
                    )
                }
            }
        }
```

**`getLastRealReviewForPrisoners` signature change:**

Previously returned `Map<String, IncentiveReview>` (keyed by prisoner number). Now returns `Map<Long, LocalDateTime?>` (keyed by `bookingId`, with a nullable date for prisoners who have no real review on record). `truncateReviewDate` updated accordingly to accept a nullable input, defaulting to the 3-month lookback when `null`.

---

### 3. Prison API cleanup

Removed from `PrisonApiService`:
- `retrieveCaseNoteCountsByFromDate` method
- All related imports (`bodyToFlow`, `CaseNoteUsageTypesRequest`, etc.)

Removed from `dto/prisonapi/PrisonAPI.kt`:
- `PrisonerCaseNoteByTypeSubType` data class
- `CaseNoteUsageTypesRequest` data class
- `BookingFromDatePair` data class
- `PrisonerInfo` data class (no longer needed)

The `PrisonerCaseNoteByTypeSubType` DTO remains in use but now lives in `BehaviourService.kt` (local to the service layer where it is produced and consumed).

---

#### Spring WebClient bean

`config/WebClientConfiguration.kt` — new `caseNoteApiClientClientCredentials` bean, configured against `api.base.url.case-notes`, using the same system client credentials and timeout as the other API clients.

#### Helm environment configuration

`API_BASE_URL_CASE_NOTES` added to all three environment value files:

| Environment | URL |
|---|---|
| dev | `https://dev.offender-case-notes.service.justice.gov.uk` |
| preprod | `https://preprod.offender-case-notes.service.justice.gov.uk` |
| prod | `https://offender-case-notes.service.justice.gov.uk` |

#### Docker Compose (local development)

`docker-compose.yml` — new `case-notes-api` service added using `ghcr.io/ministryofjustice/offender-case-notes:latest`, running on port `8086`, with `API_BASE_URL_CASE_NOTES=http://case-notes-api:8080` wired into the main application container.

---

#### `CaseNotesApiMockServer` (new)

`src/test/kotlin/.../integration/wiremock/CaseNotesApiMockServer.kt`

A new WireMock server on port `8096` replacing `PrisonApiMockServer.stubCaseNoteSummary()` across all integration tests.

**Per-prisoner request body matching:**

Each prisoner's stub uses a JSONPath filter expression to match requests that include that prisoner's prison number anywhere in the `personIdentifiers` array:

```kotlin
post("/case-notes/usage")
    .withRequestBody(matchingJsonPath("$.personIdentifiers[?(@ == '$personIdentifier')]"))
```

This approach is robust to batching: it fires correctly whether the array contains one prisoner or several, and avoids the index-based `[0]` matching that would fail for multi-prisoner batched calls.

**Why per-prisoner stubs matter:**

The new API is called once per unique `from` date rather than once for all prisoners. If a single stub returns all prisoners' data for every call, and N calls are made, the transformation layer would count each prisoner's case notes N times. Per-prisoner stubs ensure each call returns only the data for the prisoners in that specific request.

**Integration test data note:** In the current integration test setup all five test prisoners have distinct effective `from` dates (recent individual review times), so each call is single-prisoner — but the filter-expression matching future-proofs the stubs against test data that would trigger real batching.

`CaseNotesApiMockServer` is started/stopped alongside the other mock servers in `IntegrationTestBase`, and its health ping stub is wired into `HealthCheckTest`.

#### `BehaviourServiceTest` updated

- Mock expectations changed from `PrisonApiService` to `CaseNotesApiService`
- Three separate per-prisoner `whenever` stubs for prisoners with the 3-month default date (110003, 110004, 110005) are **merged into a single batched stub** that expects `setOf("110003", "110004", "110005")` and returns a combined `NoteUsageResponse` containing all three prisoners' data
- `verify(caseNotesApiService, times(5))` reduced to `verify(caseNotesApiService, times(3))`, reflecting the batching optimisation

#### `application-test.yml`

`api.base.url.case-notes: http://localhost:8096` added so the Spring context wires the test WebClient against the WireMock server.

#### `IncentiveReviewsResourceTest`

All six call sites of `prisonApiMockServer.stubCaseNoteSummary()` replaced with `caseNotesApiMockServer.stubCaseNoteSummary()`.

---

## Testing notes

- All existing integration test assertions for `positiveBehaviours` / `negativeBehaviours` are preserved unchanged — the observable behaviour of the endpoint is identical
- The `BehaviourServiceTest` unit test verifies both the correct count aggregation and that the batching reduces the number of API calls from one-per-prisoner to one-per-unique-date
- The WireMock stubs use `NoteUsageResponse`'s `content` map structure (keyed by `personIdentifier`) matching the real Case Notes API contract


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/BehaviourService.kt
```diff
@@ -1,33 +1,68 @@
...
     val lastReviewsOrDefaultPeriods = lastRealReviews.mapValues { truncateReviewDate(it.value) }
-    val caseNoteCountsByType = getCaseNoteUsageByLastReviewDate(
-      prisonApiService.retrieveCaseNoteCountsByFromDate(
-        behaviourCaseNoteMap.keys.toList(),
-        lastReviewsOrDefaultPeriods,
-      ).toList(),
-    )
+
+    val personIdentifierToBookingId = reviews.groupBy { it.prisonerNumber }
+      .mapValues { review -> review.value.first().bookingId }
```
xoen, line 27: This mapping assumes that 1 prisoner number matches to 1 bookingId. Potentially a prisoner number can have multiple bookingIds but maybe in practice this is fine?

And I see this case is covered by the error in line 37: `error("No prisonerNumber found for bookingId: $bookingId")`

xoen (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/BehaviourService.kt
Mjwillis: For a current prisoner in prison there is a 1:1 relationship between number and bookingId

PR #776, opened 2026-06-03
IR-1662 Upgrade dependencies and modernise test containers 

Mjwillis:
## Summary

Bumps project dependencies to their latest versions (via the `useLatestVersions` Gradle task)
and modernises the Testcontainers-based integration test setup, removing usages of deprecated
Testcontainers APIs. Includes the fixes needed to keep the build green after the upgrade.

## Dependency changes (`build.gradle.kts`)

- `uk.gov.justice.hmpps.gradle-spring-boot` plugin `10.2.3` → `10.3.1` (transitively updates the
  Spring Boot BOM and BOM-managed libraries such as flyway, jackson, commons-lang3)
- `hmpps-kotlin-spring-boot-starter` `2.2.0` → `2.5.0`
- `hmpps-sqs-spring-boot-starter` `7.3.1` → `7.3.2`
- `opentelemetry-extension-kotlin` `1.61.0` → `1.62.0`
- `opentelemetry-instrumentation-annotations` `2.27.0` → `2.28.1`

`kotlinx-coroutines-core` / `-reactor` were **left at `1.10.2`** (the version the Spring Boot BOM
manages). `useLatestVersions` proposed `1.11.0`, but the Spring Boot platform forces
`coroutines-core` back to its managed version via a dependency constraint while leaving
`coroutines-reactor` floating, producing a version split. At runtime this threw
`NoSuchMethodError: kotlinx.coroutines.Job.cancel$default(...)` from
`MonoCoroutines.dispose()` on the reactive request path. Keeping both pins aligned with the BOM
avoids fighting Spring Boot's dependency management.

## Test infrastructure changes

- Replaced the generic `Testcontainer` base class with dedicated `PostgresContainer` and
  `LocalStackContainer` objects, switching to the non-deprecated Testcontainers APIs
  (`org.testcontainers.postgresql.PostgreSQLContainer`,
  `org.testcontainers.localstack.LocalStackContainer`).
- `TestBase`: the R2DBC URL is now supplied via a lambda that converts the JDBC scheme
  (`jdbc:` → `r2dbc:`); previously the JDBC URL was being registered against `spring.r2dbc.url`.
- `LocalStackContainer`: set `DEFAULT_REGION=eu-west-2` on the container and register
  `hmpps.sqs.region`. The configured domain-events topic ARN is region-pinned to `eu-west-2`, so
  without these the SNS topic and SQS queues are created in different regions and domain-event
  messages never route to the listener.
- `SqsIntegrationTestBase.cleanQueue`: extracted a `HmppsQueue.purgeAndAwaitEmpty()` helper and
  increased the drain await to 30s. `PurgeQueue` is asynchronous and SQS message counts are only
  approximate, so the previous implicit 10s await flaked when a preceding test left a message
  mid-consumption.

## Testing

- `./gradlew test` — full suite green (288 tests).
- `./gradlew ktlintCheck` — passes.
  </content>


cugorji-justice (approved):
LGTM

PR #786, opened 2026-07-31
IR-1808 Handle READMISSION_SWITCH_BOOKING to correct incentive levels

Mjwillis:
[IR-1808](https://dsdmoj.atlassian.net/browse/IR-1808)

## Problem

22 prisoners in production hold a different incentive level in DPS than in NOMIS — DPS shows the receiving prison's default (typically Standard) where NOMIS correctly shows the level held before release (typically Enhanced).

The sequence:

1. A prisoner is recalled on licence and NOMIS staff mistakenly admit them on a **new** booking.
2. prisoner-search emits `prisoner.received` / `NEW_ADMISSION`, and we write a review against that booking at the prison's `defaultOnAdmission` level.
3. Staff spot the error, release the prisoner and re-admit them onto the **original** booking. NOMIS is back to Enhanced.
4. prisoner-search emits `prisoner.received` / `READMISSION_SWITCH_BOOKING` — which we ignored, falling through to the `else` branch of `processOffenderEvent`.

Nothing ever reconciled the two afterwards.

### Why the wrong level wins

`IncentiveStoreService.saveIncentiveReview` clears `current` via `updateIncentivesToNotCurrentForBooking(bookingId)`, which is booking-scoped, and the DB agrees — `V1_14` creates `one_booking_id_with_current_true_idx ON prisoner_iep_level (booking_id) WHERE current = true`. Uniqueness is per *booking*, not per *prisoner*.

So the prisoner ends up with two `current = true` rows on two bookings. Prisoner-scoped reads (`getPrisonerIncentiveHistory(prisonerNumber)`, `currentReviewSnapshotFor`, `SubjectAccessRequestService`) break the tie by latest `review_time` — and the erroneous Standard review is newer, so it wins.

## Change

`READMISSION_SWITCH_BOOKING` now routes to `reinstateReviewsAfterBookingSwitch`, which:

- resolves the reinstated booking from `PrisonerSearchService.getPrisonerInfo` (the event carries only `nomsNumber`, `reason` and `prisonId` — no booking IDs at all);
- stands down the review this service authored against the mistaken booking;
- reinstates the latest review on the reinstated booking if it is no longer current;
- recomputes the next review date, publishing its events after the transaction commits;
- publishes `incentives.iep-review.updated` via the existing `publishIfCurrentReviewChanged` (from IR-1713) so prisoner-search refreshes;
- records an `AuditType.PRISONER_BOOKING_SWITCHED` audit event.

Reviews are flipped, never deleted, so the prisoner's history stays intact.

### Identifying the mistaken booking

prisoner-search only emits this reason when the prisoner's current booking is **not** their highest-numbered one (`PrisonerMovementsEventService.isNotLatestOf`), so the mistaken booking is always the newer one.

The handler takes the **highest** booking above the reinstated one, not all higher bookings. That distinction matters when a prisoner is switched back *past* an intermediate booking: with bookings 100 / 200 / 300, a switch from 300 back to 100 would otherwise wrongly stand down a legitimate `current` review on 200. Those older rows stay current because nothing ever clears them, and they are valid history from previous sentences. Covered by a dedicated test.

Reviews authored by a human (`reviewed_by != 'INCENTIVES_API'`) are never stood down.

## On reading the booking from prisoner-search

prisoner-search is a cache of NOMIS driven by the same events, so there is an obvious question about whether it could still hold the stale booking when we call back. It cannot, for two independent reasons:

- **Ordering** — the index write precedes event generation synchronously: `PrisonerSynchroniserService.doReindexUpdate` calls `prisonerRepository.updatePrisoner(...)` and only then `generateAnyEvents(...)`.
- **Publish delay** — events go via an internal queue with `publish.delayInSeconds: 5` in production, comfortably past OpenSearch's ~1s refresh interval.

And the failure mode is benign regardless: a stale read would return the mistaken, *higher* booking, so nothing matches the predicate and the guard short-circuits to a no-op. A stale read can cause a missed fix, never a wrong one.

Reading the booking from prison-api instead was considered and discounted — see the ticket for the detail.

## Testing

- 7 unit tests in `PrisonerIncentiveReviewServiceTest` covering reinstatement, the intermediate-booking case, human-authored reviews, the no-op path, event publication and a null `nomsNumber`.
- End-to-end test in `PrisonOffenderEventListenerIntTest` publishing the message onto the queue and asserting the `current` flags swap and `incentives.iep-review.updated` is emitted.
- `./gradlew build` green — 300 tests.

## Scope

This PR stops the problem recurring. The one-off correction for the 22 prisoners already affected will follow as a separate change — it needs a `V1_35` migration with explicit review IDs derived from production, plus a decision on how to refresh prisoner-search afterwards, since a migration writes to the DB without emitting domain events. Both are tracked as open questions on IR-1808.

Deploying this on its own is safe: with no `READMISSION_SWITCH_BOOKING` event arriving for an already-affected prisoner, the handler simply never fires for them and their data is untouched.

[IR-1808]: https://dsdmoj.atlassian.net/browse/IR-1808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonerIncentiveReviewService.kt
```diff
@@ -334,6 +335,75 @@ class PrisonerIncentiveReviewService(
...
+   * The `NEW_ADMISSION`/`READMISSION` event for the mistaken booking will already have written a
+   * default-level review against it. Because `current` is unique only per booking, that review is
+   * still current and, being the most recent by review time, wins every prisoner-scoped read —
+   * masking the level the prisoner actually holds on the reinstated booking.
+   *
+   * Reviews are only flipped, never deleted, so the prisoner's history stays intact.
+   */
+  private suspend fun reinstateReviewsAfterBookingSwitch(prisonOffenderEvent: HMPPSDomainEvent) {
+    val prisonerNumber = prisonOffenderEvent.additionalInformation?.nomsNumber ?: run {
+      log.warn("prisonerNumber null for prisonOffenderEvent: $prisonOffenderEvent")
```
petergphillips, line 355: shouldn't this be an exception so that it goes on the dead letter queue and can be investigated? Can't see why it would happen anyway tbh

petergphillips (approved):
LGTM

cugorji-justice (approved):
LGTM

PR #787, opened 2026-08-03
IR-1808 Add endpoint to repair prisoners left on a switched-away booking

Mjwillis:
[IR-1808](https://dsdmoj.atlassian.net/browse/IR-1808)

Follows #786.

## Why

#786 handles the `READMISSION_SWITCH_BOOKING` event so the problem stops recurring. It does nothing for the prisoners already affected.

The incentives reconciliation report in hmpps-prisoner-to-nomis-update still shows **21 mismatches** as at 03/08/2026, **four of them new** since the 27/07 run — the problem was still actively producing new cases while #786 was being written. (Five others self-resolved in the same week, presumably because a real review touched them.)

## Change

```
POST /incentive-reviews/prisoner/{prisonerNumber}/repair-booking-switch[?dry-run=true]
```

`ROLE_INCENTIVE_REVIEWS` + write scope. It runs the same correction the event triggers — the event handler's body is extracted into a shared method rather than reimplemented, so the two cannot drift apart.

The response reports what changed, so a run can be checked against NOMIS afterwards:

```json
{ "prisonerNumber": "A9436AA", "bookingId": 3064398, "outcome": "REPAIRED",
  "dryRun": false, "levelCodeBefore": "STD", "levelCodeAfter": "ENH",
  "reviewIdsStoodDown": [2345], "reviewIdReinstated": 1234 }
```

- `?dry-run=true` returns the same body without writing or publishing anything.
- Repairing a prisoner who is already correct returns `NOTHING_TO_DO` and touches nothing, so the list is safe to re-run.
- Audited as the calling user rather than `INCENTIVES_API`, since a person asked for it.

## Why an endpoint rather than a Flyway migration

The original plan on the ticket was `V1_35__fix_readmission_switch_booking.sql` with review IDs derived from production. A migration writes to the database without emitting domain events, so prisoner-search would have carried on serving the stale level until something else touched each prisoner — and the ticket carried an open question about replaying prisoner-search's `PUT /events/prisoner/received/{prisonerNumber}` admin endpoint to work around that.

Going through the service instead means the repair publishes `incentives.iep-review.updated` itself. That closes the open question, and avoids hard-coding production review IDs into a migration that would match nothing in dev or preprod. The migration approach is dropped.

Worth noting for the sync side: hmpps-prisoner-to-nomis-update subscribes only to `incentives.iep-review.inserted` (`IncentivesDomainEventListener.kt:35`), not `updated`, so the repair will **not** write anything back into NOMIS. That is the desired behaviour — NOMIS is already correct; it is DPS that is wrong.

## Running it

Dry-run the full list first. Anything returning `NOTHING_TO_DO` is mismatching for some other reason and needs investigating separately rather than being assumed fixed. Then re-run without `dry-run` and confirm against the next reconciliation report.

## Testing

- 4 unit tests in `PrisonerIncentiveReviewServiceTest` — repair, event publication and audit actor, dry run, and the already-correct no-op.
- 5 tests in `ManageIncentiveReviewsResourceTest` — 401, 403 without write scope, the repair end to end, idempotency on a second call, and dry run leaving the rows untouched.
- `./gradlew build` green — 309 tests.

[IR-1808]: https://dsdmoj.atlassian.net/browse/IR-1808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/BookingSwitchRepair.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/dto/BookingSwitchRepair.kt
@@ -1,0 +1,57 @@
+package uk.gov.justice.digital.hmpps.incentivesapi.dto
+
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "What repairing a prisoner after a booking switch did, or would do in a dry run")
+enum class BookingSwitchRepairOutcome {
+  @Schema(description = "The prisoner's current incentive level was moved back onto the reinstated booking")
+  REPAIRED,
+
+  @Schema(description = "The prisoner's reviews were already correct, so nothing was changed")
+  NOTHING_TO_DO,
+}
+
+@Schema(
+  description = "Outcome of repairing a prisoner whose incentive level was left on a booking that NOMIS has " +
+    "since switched away from",
+)
+data class BookingSwitchRepairResult(
+  @param:Schema(description = "Prisoner number", example = "A1234BC")
+  val prisonerNumber: String,
+
+  @param:Schema(
+    description = "The booking prisoner-search reports the prisoner is on — the one their level is reinstated to",
+    example = "1234567",
+  )
+  val bookingId: Long,
+
+  @param:Schema(description = "What was done")
+  val outcome: BookingSwitchRepairOutcome,
+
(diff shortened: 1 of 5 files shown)


jonesj2193 (approved):
From what I can tell it looks good.

petergphillips (approved):
LGTM

